PR #46, opened 2015-08-17
Update Smack entry and remove aSmack's

Flowdalic:
aSmack is depreacted and "replaced" by Smack.


imaginator:
Thanks for the update.

PR #62, opened 2015-10-22
Remove the fixed width in logos.sass

xnyhps:
Not all logos have a 1:1 ratio. The Apple logo is looking quite bad, because it is 748x909. Removing the width but leaving the height makes them all look consistent.


imaginator:
Hey @xnyhps - perhaps this is a case for pre-cropping logos we upload. Can you give me an idea of where there is a problem?

xnyhps:
Right there on the main page:

screen shot 2015-10-22 at 21 55 10

PR #64, opened 2015-10-26
Also update the static app.css file with the logo change from #62.

xnyhps:
The current change isn't working, I had forgotten to run `grunt`.


xnyhps:
Okay, I'm sorry, this is getting embarrassing, the change is still not there. There were other changes to the app.css file (and looking at diffs for lines that are longer than fit on my screen vertically is a pain), so I didn't check.

imaginator:
Thanks for working so hard to get this right!

PR #82, opened 2015-11-09
Tentative build changes to also build XEPs

imaginator:
Looks like building XEPs will be easy... famous last words?


Makefile
```diff
@@ -104,6 +104,7 @@ cf_upload: publish
 github: publish
```
SamWhited: This should probably be marked as `.PHONY`

PR #87, opened 2015-11-10
Master

PeterWaher:
Provided Definition, Introduction, Outline & info on bindings.


imaginator:
Hi Peter, 

Nice changes thanks for looking this. 

I suggest that we keep the page to a brief introduction and then split out details like connection binding and other aspects to deeper level pages or point over to the wiki with a "learn more" link. We want to avoid hitting someone with a wall of text and keep the tone to a light "XMPP can help you solve your IoT problems" two or three paragraphs. 

To give you and idea, the other pages in this category try to just give a high level overview for someone not familiar with XMPP.

- http://new.xmpp.org/uses/instant-messaging
- http://new.xmpp.org/uses/social

And then provide links for someone who wants to explore more.

PeterWaher:
Ok. Thanks for input. It might be good to provide some overview with links. I'll continue to write in this one page document for now, just to keep everything in one place while writing, and when I get the disposition more or less sorted out, I can break text out into sub-pages. Is that OK? It's also easier to proofread that way, and to maintain an overview. But you're correct that a short introductory page as a first page is better.

imaginator:
Sub-pages would be great!

PR #95, opened 2015-11-12
Master

PeterWaher:
I have broken the IoT page into several pages, leaving a short introduction with links to underlying sub-pages.


imaginator:
Some broken links for example: http://new.xmpp.org/uses/iot/provisioning.md#automatic-decision-making

PR #104, opened 2015-11-17
Clean and complete the clients list

edhelas:
I've put my own project (Movim) in the list and I made a big cleaning of the list (check the second commit description for more details)


weiss:
:+1:

Personally I'd opt for removing most clients you [mentioned in the second commit message][1] (maybe with the exception of [Coccinella][2], due to the special feature it provides).  I don't think keeping them on the list helps (new) users.

[1]: https://github.com/edhelas/xmpp.org/commit/b9a75a5f718a13117447ab742bf59d65fe7eefa0
[2]: http://thecoccinella.org/

imaginator:
+1 on removing really old clients. If the authors are still looking after them, we can easily accept a new PR for them.

PR #105, opened 2015-11-19
Remove some old and non updated clients

edhelas:
I have removed all the clients that I found that didn't had updates since 2012


imaginator:
Thanks!

PR #112, opened 2016-01-26
Minor structural fixups

intosi:
These came out of my attempt to reduce the number of broken references. Two pages were missing <a name> tags where they actually needed them.  The membership page was slightly underformatted, leading to poor rendered result.


bear:
+1

PR #114, opened 2016-01-29
Add Erlang logo

erszcz:
After all, ejabberd and MongooseIM are quite well known in the community. What a pity it would be if they weren't represented on this page, wouldn't it? ;)


michalwski:
:+1:

PR #119, opened 2016-02-01
Fix local links

intosi:
A lot of local links pointed either back to http://, or occasionally to
a path without a leading /.


imaginator:
Yay - Thanks!

PR #120, opened 2016-02-01
Add security notices

intosi:
Also, a theme improvement (I hope)


intosi:
Please review the theme change more carefully than regular content changes.

bear:
oh bless you sir - if i'm reading this properly we can remove the sidebar def's from child page .md files?

+1 if so

intosi:
That's the idea.  The security notices themselves do this already.  It's not too sophisticated, so beware of dragons, and existing content hasn't been updated yet, but it does seem to work.

It also automates highlighting of the top navigation section for pages deeper than one level.

bear:
cool, then let's merge this in and we can backfill the existing pages as we touch them

PR #122, opened 2016-02-03
two new members elected

xdumaine:
http://wiki.xmpp.org/web/Meeting-Minutes-2016-02-02


bear:
+1 - thanks for the update

PR #124, opened 2016-02-03
Add page: /about/the-jabber-project

intosi:
Previously missing, linked from on the /about/history page.


intosi:
Obviously missing page was missing, adding back

PR #125, opened 2016-02-03
Add missing pages

intosi:
* Council policies and procedures
* XSF IPR Policy


intosi:
More missing content, pushing it myself.

PR #137, opened 2016-02-16
Restored (most) library URLs

guusdk:
Most of the URLs for the libraries were broken (it appears that an import script
failed). In this commit, I:

- manually fixed most of them;
- replaced Babylon with Skates;
- removed links to project I could not resolve with a quick Google;
- re-indented the table.


imaginator:
thanks @guusdk

PR #138, opened 2016-02-16
Fix links of those that can answer questions

guusdk:
When not providing a protocol, markdown resolves a link as a relative one.
I think both addresses double as a JID and email address. In the spirit of
the website, I opted for the former.


bear:
thanks!

PR #142, opened 2016-02-23
Adding Sharp.Xmpp Library

pgstath:
Added Sharp.Xmpp C# library


bear:
thanks for the update!

PR #144, opened 2016-02-23
Add the xmpp-client by @agl to the clients list

felipedau:
I noticed https://github.com/agl/xmpp-client was not on the list and I felt that it should be. It is a terminal application with a simple and friendly interface.


bear:
thanks

felipedau:
Thanks for merging it. Forgot to mention that it also supports Tor!

PR #147, opened 2016-03-04
Remove iOS Messages from list of supported clients

wfdd:
The iOS Messages app does not support the XMPP protocol.


bear:
thanks for the fix

PR #150, opened 2016-03-14
Add sortable.js to extensions table

### LINKED ISSUE #103: Extensions only sortable on name
Would be nice if the extensions could be sorted on any column, like the old page.

lloydwatkin:
Fixes #103


bear:
nice!

PR #153, opened 2016-03-16
Fix UTF-8 encoding of ®

xnyhps:
Replace occurrence of REPLACEMENT CHARACTER with REGISTERED SIGN on the Jabber trademark page.


intosi:
Thanks.

PR #156, opened 2016-03-18
Fix link to wiki page.

guusdk:
Without an explicit scheme definition, the website attempts to resolve links
relative instead of as an absolute path.


intosi:
Merged. Thanks!

PR #181, opened 2016-05-17
Add more error checking to buildCompleteWebsite.sh

tfar:
This script should fail, i.e. return non-zero, if some part of it fails.

### DIFF
--- a/buildCompleteWebsite.sh
+++ b/buildCompleteWebsite.sh
@@ -20,6 +20,12 @@ OUTPUTDIR=$BASEDIR/output
 CONFFILE=$BASEDIR/pelicanconf.py
 PUBLISHCONF=$BASEDIR/publishconf.py
 
+function error_handler {
+  rm -rf "$OUTPUTDIR"
+  exit 1;
+}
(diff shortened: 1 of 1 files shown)


buildCompleteWebsite.sh
```diff
@@ -26,31 +26,31 @@ mkdir -p build
 cd build
 if [ -d "xeps" ]; then
 	cd xeps
-	git pull
+	git pull || exit 1;
```
bear: I believe you can add `set -e` to the beginning of the script to force and exit for any non-zero result

tfar: I've heard `set -e` breaks if you use if-statements that are based on exit code of programs. However I'm very bad at bash scripting.

bear: It can, but IMO if your doing that inside of a CI script then you have other issues.

For that pattern then `trap "exit 1" EXIT` works

tfar:
@bear Does the new version look better?

bear:
a thing of beauty - thanks for the adjustment!

:+1:

Issue #186, opened 2016-05-23
Add robots.txt


guusdk:
To give instructions to robots crawling the website, we should add a robots.txt. See http://www.robotstxt.org/

We currently don't serve the file, which causes a lot of 404s to be logged. That's easily resolved by adding a file that allows all access. When needed, we can improve on that.

PR #187, opened 2016-05-23
#186: Add an permits-all robots.txt file.

guusdk:


bear:
@guusdk you need to add any static file to the config file so pelican knows to copy it:

https://github.com/xsf/xmpp.org/blob/master/pelicanconf.py#L21

guusdk:
Like that?

bear:
yep!

PR #193, opened 2016-05-31
Arranged clients by operating system

conorohiggins:
I have rearranged the list from an alphabetic list of all the clients, to a list categorized by operating system.

I think this is more intuitive, and will make it easier for people to get started with XMPP, as they are coming from an existing operating system and looking for a client to fit that.

### DIFF
--- a/content/pages/software/clients.md
+++ b/content/pages/software/clients.md
@@ -26,75 +26,79 @@ An XMPP client is any software or application that enables you to connect to an 
 
 __See something missing?__ Any list of XMPP servers, clients or libraries will, due to the dynamic and evolving nature of the XMPP market, be out of date almost as soon as it’s published. If you spot mistakes, errors or omissions in the table below, please [submit a pull request!](https://github.com/xsf/xmpp.org)
 
-| Name                           | Platform(s)                                           |
-|--------------------------------|-------------------------------------------------------|
-| [Adium](http://adium.im)                          | OSX                                                |
-| [AQQ](http://aqq.eu)                            | Windows                                            |
-| [AstraChat](http://astrachat.com)                      | Mobile (Android, iOS) / Linux / OSX / Windows                            |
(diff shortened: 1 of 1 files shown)


guusdk:
Thanks for your contribution! 

I do wonder if this layout makes it hard to add a new client, in the future. Say that you need to add a new client somewhere in the top of the OS X column: you'd have to re-arrange the that column, but leave the other ones in place. Is there an easy way of doing that?

I don't particularly like the duplicate entries, but it doesn't bother me to much either. The "misc"  section is odd though. Those entries almost feel "left out".

conorohiggins:
The Misc section seemed weird to me too. It's not the formatting that makes
it weird; what the hell is Asha?

Your point about rearranging columns is valid. I can't think of a way to
automatically 'push' entries down a column in markdown.

guusdk:
@conorohiggins How would you like to proceed? The rearranging complexity is a deal-breaker for me, as it'll seriously hamper future additions to this page (but then again, no-one died and made me king of anything).

Perhaps leaving the table as-is, but making it sortable, and/or apply a filter, would bring a similar functional improvement. Both sorting as well as filtering is already in place with the table at http://xmpp.org/extensions/index.html - so that should be do-able on this page too.

conorohiggins:
I think we should proceed by adding filtering, so people can find one
that fits their operating system. I'll look at how that's done on the
extensions page you linked to, and replicate that.

guusdk:
@conorohiggins *poke* ;)

conorohiggins:
What do we think of making a sortable Javascript table like the one at https://github.com/xsf/xmpp.org/blob/master/content/pages/extensions.html

guusdk:
Go for it!

horazont:
I think we can close this, the client listing has been completely overhauled. Grouping it by OS will be more trivial once the renewal period is over, and then we can think of nice UI. In any case, it’ll need a new PR.

PR #197, opened 2016-06-11
Replaced the Go library

zach-klippenstein:
The current Go library is marked as deprecated. I updated the table to point to the one referenced in the deprecation notice.


SamWhited:
Is there a canonical import path for this (or somewhere else where the docs can be easily generated and that `go get` knows about)? If so, I'd recommend linking to that (or the godoc.org page for it) instead.

bear:
oops, great catch - sorry for merging quickly

zach-klippenstein:
@SamWhited Should I change it to cjones.org/hg/go-xmpp2.hg/xmpp, or do you want to, or is it ok as-is?

SamWhited:
@zach-klippenstein That doesn't appear to resolve in a browser (or be usable from godoc.org); I'm not sure what the best thing to do here is.

PR #213, opened 2016-10-10
Change the link to the specific page

wilsonmar:
So people can get right to work rather than having to hunt around.

Just an idea. Use it or not.


guusdk:
Good idea - this should make it even easier to add stuff. Thanks.

PR #214, opened 2016-10-15
Add Slixmpp to the list of libraries

linkmauve:
It is a friendly fork of SleekXMPP, replacing threading with asyncio and thus requiring Python 3.4+.


bear:
+1

PR #216, opened 2016-10-24
libraries: remove go-xmpp2

SamWhited:
It is unmaintained and not actually compatible with any of the standard
Go tooling.

That being said, it probably does work; is that enough for us to list it?


bear:
maybe someone will see it and want to fix it... so +1

PR #222, opened 2016-11-25
Add council bio

SamWhited:
Added a bio. These things are hard.


guusdk:
You did it like a pro though.

PR #223, opened 2016-11-25
Update the editors list

SamWhited:
Update list of editors.

Also linked the mailing list address and removed some insignificant trailing whitespace.


guusdk:
For future reference: the member list at http://xmpp.org/about/xsf/members.html lists each role of a member. When you update a page like the editors list, there's probably a change needed in the member list as well.

PR #230, opened 2016-12-03
libraries: update Loudmouth information

mhallendal:
Updated the C library which is now being maintained by the MCabber community and removed the Ruby bindings that are unmaintained.

Note, I'm the original author but @McKael is currently the maintainer.


intosi:
Thanks, Mikael.

PR #231, opened 2016-12-08
Update of compliance suites in extensions page

Nyco:
From 2009 to 2016.


SamWhited:
The 2016 suites aren't really official yet and there are still some disagreements between me and everyone else which I'm not 100% sure what to do about. Chances are they'll end up being bumped into 2017.

guusdk:
What shall we do here? Close and open a new one after the 2017 specs have been finalized?

guusdk:
Closed in favor of #249

PR #236, opened 2016-12-30
Update link to Apache Vysper

RubyTuesdayDONO:
Updated the Apache Vysper link to point directly to the Vysper page within its enclosing project (MINA).

Other rows in the table were only updated with padding to align columns.


SamWhited:
Apache Vysper doesn't seem to have been updated in several years and is implementing a very old version of the core protocol. Do we still want to link to it at all?

guusdk:
Perhaps we should delete it completely, but as it is, this PR is an improvement. Let's follow up with a new PR if/when desired.

PR #237, opened 2017-01-02
Remove link to non-existent all.shtml

Zash:
> A [list of all XEPs](http://xmpp.org/extensions/all.shtml) (including retracted, rejected, deprecated, and obsolete XEPs) is also available.

This text is wrong, since extensions.html lists all those XEPs already, depending on the checkboxes.

### DIFF
--- a/content/pages/extensions.html
+++ b/content/pages/extensions.html
@@ -54,9 +54,8 @@
             <a href='https://datatracker.ietf.org/doc/rfc7395/'>RFC 7395</a>).
         </p>
         <p>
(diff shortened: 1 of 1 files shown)


Zash:
The sentence just before the removed one is also wrong FWIW.

tfar:
> The sentence just before the removed one is also wrong FWIW.

But not fixed by your PR?

Zash:
> But not fixed by your PR?

No, wasn't as sure about good wording for that.

PR #246, opened 2017-01-16
Website should have a 'getting started' page

guusdk:
This pull request introduces a new page that guides people that are completely new to XMPP.

This pull request results from a discussion on the XSF MUC earlier today. I thought it'd be a good idea to float the idea in a tangible way. Obviously, the commit needs work: the list of clients and services needs to be completed, and the entire thing could be made a lot prettier too.

### DIFF
--- a/content/pages/uses/getting-started.md
+++ b/content/pages/uses/getting-started.md
@@ -1,0 +1,34 @@
+---
+Title: Getting Started
+Url: uses/getting-started
+Save_as: uses/getting-started.html
+Top_menu_show: true
+Top_Menu_order: 6
+Dropdown_menu_show: false
+Footer_show: false
+Content_layout: multiple-columns
+---
+
+Do you want to start chatting with XMPP? This page helps you get up and running in a matter of minutes.
+
+# 1. Download a client
+
+There are [many, many XMPP clients](/software/clients) for you to choose from. To get you started, here are some of the more popular ones:
+
+* [Conversations](https://play.google.com/store/apps/details?id=eu.siacs.conversations&referrer=utm_source%3Dxmpp.org) (Android)
+* [Swift.IM](http://swift.im/swift.html) (OS X, Windows & Linux)
+* [Lorem ipsum](http://example.org)
+* [Dolor sit amet](http://example.org)
+
+# 2. Create an account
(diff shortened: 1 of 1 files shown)


guusdk:
Here's a screenshot of the commit: https://ibin.co/3987T3lKd72z.png

(Note that locally, I've not been able to get the menu to work right)

bear:
👍  

let's get it on the site as nothing makes people update/fix/add things than seeing it on the web

guusdk:
I like the way you're thinking @bear :)

Up until now, the XSF remained neutral to a fault, where it concerns implementation or service advocacy. That changes somewhat with this PR, as it pushes forward some clients and services.

Although I believe this is acceptable (and to this degree, desirable), others have voiced concerns.

Is this PR something that we want to run by the board?

bear:
I don't think it requires that - let's ask forgiveness if someone really sees something they don't like

Kev:
I hate to be a dissenting voice here, but the website is the voice of the XSF and the XSF has previously decided to be neutral. I don't think it's up to anyone here to change that. Bring it before Board. It's either a simple 'yes', in which case no harm done by asking, or it's a 'no', in which case it clearly shouldn't have been merged.

ge0rg:
@Kev would it be sufficiently neutral to only link to the XSF client list and a public server list like https://xmpp.net/directory.php ?

In that case I'd argue that we make a first, neutral, version of the page and get that out, so that it's not blocked by the much needed discussion. Regarding the overall neutrality question, I think this is something that should be brought up at the summit, and be decided by the whole XSF.

Kev:
Referring to third-party lists doesn't seem harmful to me, if appropriately linked.
I think we should bring this PR before Board and see if we can do it all though.

guusdk:
I've herded this PR for six weeks now, but little progress is made, discussion-wise. In various discussions that started, but never concluded, the general consensus appeared to be in favor of leaving strict neutrality.  As far as I can tell, the last board meeting didn't discuss the neutrality topic, while the one before that bulldozered the issue ahead.

I'm merging this issue. If people care enough about this issue to disagree, I'm assuming that my merge will trigger a follow-up.

PR #249, opened 2017-01-28
Bye bye 2009 compliance suites~

linkmauve:


guusdk:
There's overlap between #249 and #231.

Issue #224, opened 2016-11-28
Refs missing


SamWhited:
The XEP refs that are linked from several places on various IETF sites (eg. see the "citation libraries" section here: https://xml2rfc.tools.ietf.org/) are missing:

* http://www.xmpp.org/extensions/refs/jsf-xeps.tgz
* http://www.xmpp.org/extensions/refs/jsf-xeps.zip
* http://www.xmpp.org/extensions/refs/

and this does not appear to be updating:

* http://www.xmpp.org/extensions/rss.xml

stpeter:
Aha, I missed this one.

I've been in communication with the IETF tools team and they updated the tgz and zip URLs.

We still need to fix the index file and Atom/RSS feeds.

guusdk:
Do we serve that content elsewhere now? If so, it'd be good to also have redirects in place.

Echolon:
Hi @stpeter ,

do you have an update on this?

wurstsalat3000:
This issue should be moved to https://github.com/xsf/xeps

Echolon:
> This issue should be moved to https://github.com/xsf/xeps

@horazont do you have sufficient right to move it?

Echolon:
@Zash I think this needs to be moved as well. just spottet

Zash:
@Echolon I don't have move privileges.

PR #254, opened 2017-02-07
Adding categories to blog.

guusdk:
Currently, all blog posts are in the same default category 'misc'. This PR adds some other categories.

Certain posts could go into two categories, but I've yet to figure out how that's done. Comma-separated won't work, nor will adding the header twice. 

Feel free to improve or add to this.

### DIFF
--- a/content/posts/blog/19th-xmpp-summit-and-fosdem-2016.md
+++ b/content/posts/blog/19th-xmpp-summit-and-fosdem-2016.md
@@ -3,6 +3,7 @@ title: The 19th XMPP Summit and FOSDEM 2016
 date: 2016-02-08 22:25
 authors: Tobi; Nyco
 blog_id: blog
+category: XMPP Summit
 ---
 
 ##19th XMPP Summit
--- a/content/posts/blog/2011-q4-membership-application-period-has-started.md
+++ b/content/posts/blog/2011-q4-membership-application-period-has-started.md
@@ -3,6 +3,7 @@ title: 2011 Q4 Membership Application Period has started
 date: 2011-09-22 21:47
 author: bear
 blog_id: blog
+category: XSF Organisational
 ---
 
 Alexander Gnauck has announced to the members@xmpp.org mailing list that the [membership application Wiki page](http://wiki.xmpp.org/web/Membership_Applications_Q4_2011) for the next application period is ready.  Now is the time to start looking for new members, or remind those who have dropped to reapply.
--- a/content/posts/blog/2013-annual-meeting-and-voting-results.md
+++ b/content/posts/blog/2013-annual-meeting-and-voting-results.md
@@ -3,6 +3,7 @@ title: 2013 Annual Meeting and Voting Results
 date: 2013-11-07 16:01
 author: bear
 blog_id: blog
+category: XSF Organisational
 ---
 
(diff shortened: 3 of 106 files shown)


tfar:
Many constraints in pelican are by design, so is that a post can only be in one category, see https://github.com/getpelican/pelican/issues/350 . It can have more than one tag though.

guusdk:
Ah, ok. Should we move to tags then, or is a single category good enough?

tfar:
I imagine there are use case for *tags* and *categories*. With tags you can have multiple and categories can build hierarchies I think. I'd go with only tags as we probably won't need a hierarchy in our blog posts and tags provide more flexibility.

guusdk:
Meh. When I switch 'categorie' for 'tags', then nothing visually changes (as compared to before I started this PR). When I use 'tag' (instead of 'tags'), Pelican throws this error during page generation:

```CRITICAL: AttributeError: 'unicode' object has no attribute 'slug'
Traceback (most recent call last):
  File "/usr/local/bin/pelican", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python2.7/dist-packages/pelican/__init__.py", line 465, in main
    pelican.run()
  File "/usr/local/lib/python2.7/dist-packages/pelican/__init__.py", line 179, in run
    p.generate_output(writer)
  File "/usr/local/lib/python2.7/dist-packages/pelican/generators.py", line 599, in generate_output
    self.generate_feeds(writer)
  File "/usr/local/lib/python2.7/dist-packages/pelican/generators.py", line 300, in generate_feeds
    self.settings['FEED_ALL_ATOM'])
  File "/usr/local/lib/python2.7/dist-packages/pelican/writers.py", line 123, in write_feed
    self._add_item_to_the_feed(feed, elements[i])
  File "/usr/local/lib/python2.7/dist-packages/pelican/writers.py", line 49, in _add_item_to_the_feed
    link = '%s/%s' % (self.site_url, item.url)
  File "/usr/local/lib/python2.7/dist-packages/pelican/contents.py", line 326, in url
    return self.get_url_setting('url')
  File "/usr/local/lib/python2.7/dist-packages/pelican/contents.py", line 188, in get_url_setting
    return self._expand_settings(key)
  File "/usr/local/lib/python2.7/dist-packages/pelican/contents.py", line 182, in _expand_settings
    return self.settings[fq_key].format(**self.url_format)
  File "/usr/local/lib/python2.7/dist-packages/pelican/contents.py", line 175, in url_format
    'tag': self.tag.slug if hasattr(self, 'tag') else '',
AttributeError: 'unicode' object has no attribute 'slug'```

Not sure how to proceed.

tfar:
`tags` was probably correct. However, our theme doesn't use tags yet in the UI. If you take a look at https://github.com/xsf/xmpp.org/blob/master/xmpp.org-theme/templates/article.html it only mentions category. So it seems using `tags` would require further adjustments to our theme. Happy to go with categories instead then.

guusdk:
I tend to agree. Let's ping @bear (ping!) to see if switching from categories to tags in the theme is easy, but if not, let's go with categories.

bear:
it doesn't really matter which is used as long as the code that generates the feeds is also updated

guusdk:
Where is the code that updates the feed? Is that the same code as what @stpeter mentions is broken in #224 ?

bear:
probably - it may end up just being a config issue

guusdk:
After the merge, categories popped up, and the single blog feed that we already had did not break. I'm declaring this a success. ;)

PR #265, opened 2017-02-16
Add subscribe URL for the standards list

jcbrand:
No URL was provided, which meant the link merely pointed to the current/same page.


guusdk:
Thanks JC.

PR #267, opened 2017-02-28
Update link for Verse

Zash:
The old link pointed to an obsolete(afaik) version of Verse.
New version unfortunately doesn't have that much in the way of docs yet (that I'm aware of).


guusdk:
tx!

PR #269, opened 2017-02-28
Generate software lists dynamically via Jinja template from JSON files

horazont:
Rationale
---

As discussed e.g. in the xsf@ MUC, we want to avoid listing outdated and abandoned projects in our software listings. This PR allows to have per-project timestamps of last renewal which make projects disappear from the listing if the last renewal is older than 365 days (configurable in pelicanconf.py).

Implementation
---

The software listings have been removed from the markdown source. Instead, those pages now use a new template [1] which generates the tables from JSON data which is loaded, preprocessed and filtered in pelicanconf.py. The JSON data contains for each project an object like this:

```json
{
    "info": [
        "Python"
    ],
    "last_renewed": "2017-02-28T14:57:45",
    "name": "aioxmpp",
    "url": "https://github.com/horazont/aioxmpp"
}
```

The ``info`` key is rendered in the second column of the software table, joined together using `` / ``. ``last_renewed`` is the timestamp of last renewal in ISO timestamp format. ``url`` and ``name`` should be self-explaining.

This is the most elegant solution for pelican I could think of. It *requires re-building the website periodically*. (I do not know whether that already happens.)

Utilities
---

To update entries, there is a command-line utility based on python in ``data/update-entry.py``. In its simplest invocation it simply updates the timestamp:
```sh
./data/update-entry.py ./data/libraries.json aioxmpp
```

It has other options which can be explored with ``--help``. Most notably, the ``--ask`` option shows the changes before applying them and allows to abort the process (not that this would be totally necessary considering that all data is checked into git).

I also put the two scripts I used to convert the markdown tables to JSON in ``data``. Use them like this:
```sh
cd data
./table-to-tuples.sed foo_raw | ./tuples-to-json.py > foo.json
```
``foo_raw`` must contain the *rows* (only from the body, not the header!) of the markdown table to convert.

Caveats
---

This requires re-building the website periodically. Not sure if this is already happening.

**Footnotes:**
1. This breaks with non-XSF themes.


guusdk:
Thanks for this!

Whenever a PR is merged that modifies one of the new JSON files, the website gets re-deployed. I'm assuming that this does all of the rebuilding that is required.

tfar:
Assuming? I thought you tested this and reviewed it before merging?

tfar:
I really would have liked to reviewed this before being merged.

guusdk:
I've reviewed it in so much that it builds locally, and gives me a populated list of software. I am not sure if whatever magic deploys the website to production from github invokes the same build process as what I do locally though.

As for reviewing: it was up for two days, it was discussed on in the chatroom, and you did not leave a mark here indicating that you wanted to review it, before a merge occurs. That being said: the merge is not the end of the world: It is not deployed to production yet (there's still time), and even if a problem would occur, we can undo and/or correct changes with follow-up commits.

Kev:
FWIW, anything that introduces new scripts that are run on the server absolutely MUST have careful review before it's merged and run on the server. If this PR hasn't had that level of review yet, please revert.

horazont:
@guusdk 
> Whenever a PR is merged that modifies one of the new JSON files, the website gets re-deployed. I'm assuming that this does all of the rebuilding that is required.

This is, in theory, *not* correct! It must be re-built periodically to evict software from the lists which has not been renewed. In practice, pull requests may come in often enough to ensure that this happens, but it needs to be considered.

guusdk:
@horazont Although I'l admit that I did not consider that at the time, I think that we should be fine with piggy-backing on all generic/unrelated changes made to the website. If that does not occur frequently enough for that list to be cleaned up periodically, we have bigger problems than an out-of-date list.

As for the reviewing: I'm happy with the changes, but yes, I am fallible and a bug might have slipped by. Feel free to revert the changes if you feel that there's a need to - but if you do so, please have some kind of procedure put in place for people to follow when considering PRs for merger. As long as that's not there, I'll continue to use my best judgement - which might or might not be regarded as good judgement by others.

PR #271, opened 2017-03-07
Fix pull request links for new data-driven software lists.

kousu:
Lists added in d8cec85 .. e6c19ce, but these links were forgotten.

This actually changes the servers and libraries links to be a direct
edit link, the way clients was. I think that's better.


guusdk:
Thanks!

PR #273, opened 2017-03-09
Adjust size of menu for its content.

guusdk:
This likely fixes the build problem introduced in commit 83f365dc99f8a60f31ea5b524e7daafedb714916

I'm unsure why this build problem manifests on our webserver, but not on Travis, nor my local environment. Any pointers on that would be appreciated. If anything, I'd like Travis to catch mistakes like these.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ cache: apt
 notifications:
   email: false
 install:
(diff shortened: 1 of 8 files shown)


intosi:
Looks good to me, thanks!

PR #278, opened 2017-03-16
Provide a manual for the update-entry.py tool and an example commit for renewal

horazont:
This adds a README.rst to the data/ directory which shows a nice manual on how to manage your softwares entry.

Also an example commit where I renew aioxmpp with a commit message showing the renewal process in detail. I hope this is okay.

This doesn’t change any code which is run by the server or any content shown on the website (well, except that it renews aioxmpp).

### DIFF
--- a/data/README.rst
+++ b/data/README.rst
@@ -1,0 +1,109 @@
+Software Directory README
+#########################
+
+The Software Directory needs yearly renewal for software to not disappear from the listing on the website. To help software authors with this, scripts are provided which help you with creating a proper pull request.
+
+The listing is split in three categories:
+
+* **Servers** are found in the ``servers.json``,
+* **Clients** are found in the ``clients.json``, and
+* **Libraries** are found in the ``libraries.json``.
+
+The tooling is the same for all three categories. What differs is what ``info`` means. For clients and servers, this is the list of platforms a project runs on. For libraries, this is the list of languages they support.
+
+**Note:** You can of course also manually edit the JSON files if that’s what you prefer. Make sure to keep the diff minimal. All times are UTC.
+
+
+Renewing your existing entry
+============================
+
+For renewing (and changing) an existing entry, you can use the ``update-entry.py`` tool from the command line. Example::
+
+  ./update-entry clients.json Conversations
+
+The tool will ask for confirmation::
+
+  old entry:
+  {
+      "info": [
+          "Mobile (Android)"
+      ],
+      "last_renewed": null,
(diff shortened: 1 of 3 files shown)


data/README.rst
```diff
@@ -0,0 +1,107 @@
+Software Directory README
+#########################
+
+As decided by the XSF Board, the Software Directory will need yearly renewal for software to not disappear from the listing on the website. To help software authors with this, scripts are provided which help you with creating a proper pull request.
```
guusdk: I would remove "As decided by the XSF Board." It does not really matter who decided this for the README.

guusdk: The text could mention more explicitly that you don't _have_ to use the tool, but can manually edit the JSON files too. Although using the tool will safeguard against formatting errors, a simple edit of the JSON files will be preferred by some, as you can do that without executing anything.

data/README.rst
```diff
@@ -0,0 +1,107 @@
+Software Directory README
```
guusdk: The ``rst`` extension of the README file is pretty uncommon. If we use a more common extension, more clients will automagically render the content properly. I'd suggest using no extension (just ``README``) or the markdown extension (``README.md``)

horazont: Can you clarify? It renders fine on GitHub, and that’s handled server-side.

guusdk:
I've added some minor suggestions - non of which are blocking issues for me.

data/README.rst
guusdk: Clarification: this is the very first time I've seen that extension being used. :)

data/README.rst
horazont: Thanks for your feedback, will integrate that into the PR today.

data/README.rst
horazont: Do you have concerns except the readability on GitHub and in plaintext (both of which is fine with rst in my opinion)?

guusdk: I'm browsing the website source using various tools (pretty much whatever I have handy at the time.) Although I'm not sure if those tools will bork over the extension, why run the risk? Using a commonly used extension will prevent an issue, if there's one. 

If you do not agree, I wil happily shrug and move on. To me, it's a trivial issue at best. :)

horazont:
@guusdk Reworded, mind to read over it again?

guusdk:
I think it's a good improvement. I've already approved the PR, but will wait with the merge for a bit to give others the chance to provide feedback.

Issue #276, opened 2017-03-14
Update or remove Event page


guusdk:
https://xmpp.org/community/events.html contains out of date information. We should either update it (and keep it updated), or remove that page.

Nyco:
We got: https://wiki.xmpp.org/web/Meetups

Which should we maintain?

guusdk:
As far as I'm concerned, the intended target for the wiki is XSF members, while the intended target for the website is anyone interested in XMPP. That's why I'd prefer to move the meetups page from the wiki to the website.

PR #281, opened 2017-03-23
276: Remove unmaintained Events page.

guusdk:
This addresses issue #276.

When this gets merged, we should add a redirect for the removed page. I suggest a redirect to the HTML page that lists the Events calendar. The redirect would be from: ``/community/events.html`` to: https://calendar.google.com/calendar/embed?src=64v3vs15qlalgqv0j7r99ikm1c%40group.calendar.google.com


guusdk:
We now have #294! No need to remove the page!

PR #290, opened 2017-03-23
Several updates to the software registry tooling

horazont:
This includes several things discussed in the xsf@ MUC.

1. a linting tool (``lint-list.py``)
2. ``update-entry.py`` now shows the changes as a diff instead of side-by-side
3. the platforms which can be used for clients and servers are defined in ``platforms.json`` and they’re enforced by the linting tool (only for renewed entries) and ``update-entry.py``.
4. rename the ``"info"`` field to ``"platforms"`` in the JSON (which needs changes in the templates and page files). This has the advantage to be clearer in the tooling and sorting more nicely in the JSON output.

The readme has been updated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,14 @@ language: python
 cache: apt
 notifications:
   email: false
+python:
+  - 2.7
+  - 3.3
 install:
-- pip install pelican==3.3 ghp-import markdown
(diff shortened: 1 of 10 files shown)


horazont:
Cleaned up the commits after @guusdk fixed the travis. Unfortunately, the attribution to the .travis.yml changes got lost in the process. We can fix this if desired.

PR #302, opened 2017-03-24
Add ChatMe

camaran:
Add ChatMe service


guusdk:
I've verified with someone in Italy: this has the appearance of a legit service provider. I'm adding it, but will note some concerns voiced in the XSF muc yesterday about the existence of this page: it runs the risk of turning into a marketing tool for third parties.

PR #305, opened 2017-03-27
Update Isode products in software listings

tfar:


horazont:
> ERROR: entry 'Stroke': should be placed behind 'Swiften' (all entries must be ordered alphabetically by case-folded name)

Please fix this :-)

tfar:
Should be fixed now. I just reran the python tool.

PR #309, opened 2017-04-04
Update client.json

yagiza:
Add information about eyeCU client.
Update information about Psi and Sim-IM clients.

### DIFF
--- a/data/clients.json
+++ b/data/clients.json
@@ -155,6 +155,16 @@
         "url": "https://wiki.gnome.org/Apps/Empathy"
     },
     {
+        "last_renewed": "2017-04-01T10:28:44",
+        "name": "eyeCU",
+        "platforms": [
+            "Windows",
+            "OS/2",
+            "Linux"
+        ],
+        "url": "http://eyecu.ru"
+    },    
+    {
         "last_renewed": null,
         "name": "Finch",
         "platforms": [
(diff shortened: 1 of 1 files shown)


guusdk:
I remain reluctant to add OS/2 as a platform. Clients listed here only appear to support that when they're multi-platform. I'd like opinion of others, before rejecting or accepting this PR.

In the mean time, it might be good to separate this PR into two: one that updates the client registration, and one that updates the platform. That will allow us to process both PRs separately.

SamWhited:
OS/2 hasn't had a release since the early 2000's as far as I am aware; I also see no point in adding it to the list.

horazont:
@yagiza Realistically, splitting the PR into two will help at least the renewal to pass. We haven’t come to a conclusion whether we want OS/2 on the list or not yet.

yagiza:
Ok. I removed mention of "OS/2" from this PR, replacing it with "Other".

yagiza:
> OS/2 hasn't had a release since the early 2000's as far as I am aware; I also see no point in adding it to the list.
I just want to notice, that "OS/2" platform include not only IBM OS/2, which support and development ended many years ago, but also its decendants, deployed under license of IBM:

- eComStation from Serenity System International (lst release was in 2013)
- ArcaOS AKA Blue Lion (first release planned on 15.04.2017)

PR #310, opened 2017-04-14
ProcessOne renewals

f055:
Renewed: ejabberd server, OneTeam client. Removed discontinued OneTeam
for iPhone client. Added Erlang/Elixir library.


guusdk:
Thanks!

PR #311, opened 2017-04-16
OS/2 platform

yagiza:
Separate PR for adding "OS/2" platform to the list

### DIFF
--- a/data/clients.json
+++ b/data/clients.json
@@ -159,7 +159,7 @@
         "name": "eyeCU",
         "platforms": [
             "Linux",
-            "Other",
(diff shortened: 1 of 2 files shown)


guusdk:
This was initially part of #309 and was split off into a new PR to judge it on its own merits.

I'm reluctant to add OS/2 as a platform. It seems a niche, at best.

guusdk:
I'm going to close this without a merge, for the reasons stated above. Apart from you, I've heard no-one that was in favor of this, while most agree that it would be over-specific at this point, to add OS/2 as a platform.

PR #312, opened 2017-04-18
Add new client: Dino

participante0:


guusdk:
You'll probably want to set the last_renewed date - otherwise the entry will be pruned from the list of clients as soon as the purge starts.

guusdk:
Hi - that's better, but you now added renewal timestamps that are unrelated to Dino. That's the responsibility of maintainers of those software stacks (the point of having timestamps in the first place is that we want to purge entries that do not get updated by their maintainers).

participante0:
I added only timestamps of a few active (and currently mantained) software: ChatSecure, Conversations, Converse.js, and Dino.

guusdk:
Fair enough. Thanks for your contribution!

PR #315, opened 2017-04-21
Add Zom

rozzin:


guusdk:
Is this a client that you can use on other XMPP domains, apart from the ZOM one?

Also, you should add a date for the "last_renenewed" attribute.

ossguy:
@guusdk Yes, you can use it with any XMPP account.  On first launch Zom provides "Create an Account" and "Sign into an existing account" options.  If you choose the latter, you get a screen with this text: "You can use Zom with an existing XMPP/Jabber account on any server you would like."  It asks for "user@domain" and a password, with a "Sign in" button.

ossguy:
To confirm, I just added a 5222.de account to Zom and it was able to send messages fine with that account.  So it does appear to support other XMPP domains just fine.

guusdk:
Excellent, thanks. When the PR is fixed (see failing check - an attribute value is missing), we can merge this!

rozzin:
OK. I have no idea how to modifiy the pull-request; I pushed up an amended revision, though: d57dd8360b6acc66e66eb46b207a7a3eec37b89b

Sorry--I initially skipped reading the README and just followed what appeared to be the conventions elsewhere in the file (and then apparently pushed a typo--o..y): most of the existing entries have null last_renewed attribs--what's up with that?

horazont:
@rozzin Those with ``null`` are old entries before the decision to add a requirement for yearly renewal was made. Those entries will be removed from the website soon-ish.

guusdk:
Thanks!

PR #323, opened 2017-06-07
Fixed Sponsor Footer Links

leonthemisfit:
Added .html to the sponsor links in footer to fix 404 errors when following links.


leonthemisfit:
URLs on new.xmpp.org work correctly without the addition of .html to links

PR #322, opened 2017-06-06
change two items

tetrode:
The Eiffel library was labeled as PHP, a correct URL for XMPPHP is provided

PR #324, opened 2017-06-07
change of three items (retry of #322)

tetrode:
The Eiffel library was labeled as PHP, a correct URL for XMPPHP is provided, Coversant removed as the company does not exist anymore.
#322 can be discarded


guusdk:
Thank you, and thank you for persisting in your change! ;)

Issue #325, opened 2017-06-14
Dead Link: IPR Policy


stpeter:
The IPR policy has moved from https://xmpp.org/about/xsf/ipr-policy  to https://xmpp.org/about/xsf/ipr-policy.html - we need to update our pointers (e.g., from https://xmpp.org/about/standards-process.html and NOTE from all XEPs so the XSLT needs fixing too).

guusdk:
Although we should update the pointers, there might be references outside of our control. I'd like to have a forward from the old to new URL, to complement the changes to our own content. Configuring the forwards is out of my control though (although I wish it wasn't - this pops up regularly - is there anything we can improve there, @Kev ?)

Kev:
There is - we can set up the rules inside the docker container that deploys the website (which is under version control, although I'd prefer ours), instead of the host.

guusdk:
You're referring to @dwd's repo? I thought everyone agreed to move that to our repo? Can't we (someone with admin powers here) migrate that here?

guusdk:
I'm actually finding several variants of the 'old' address. Not sure if they ever worked, but perhaps we should have forwarding rules for all of these:
- /about/xsf/ipr-policy
- /about/xsf/xsf-ipr-policy/
- /about/ipr-policy
- /about-xmpp/xsf/xsf-ipr-policy/
- /extensions/ipr-policy.shtml (which already redirects to a 404 on https://xmpp.org/about/xsf/ipr-policy)

These are all intended to go to /about/xsf/ipr-policy.html - right?

Also, the references to footnotes in https://xmpp.org/about/standards-process.html link to http://xmpp.org/extensions/ipr-policy.shtml#note1 (and #note2) which redirects to https://xmpp.org/about/xsf/ipr-policy#note1 and 2 (which both 404's). The reference in that document needs to be updated (I can do that), but the redirect needs to be modified too. I'll create the required anchors ( /about/xsf/ipr-policy.html#note1 and /about/xsf/ipr-policy.html#note2 )

guusdk:
The PR above addresses the content on the public webpage - but not in the extensions.

I am starting to wonder if there's a larger problem here. The timing of this issue, as well as #321 is curious. Did these links work before the server was resurrected?

guusdk:
The missing .html issue will likely be fixed with https://github.com/dwd/xsf-web-docker/pull/1 although the issue of having multiple different URLs will likely still need some work. I'll redo my commit later tonight.

[...]

PR #326, opened 2017-06-14
Fix for #325 Dead Link: IPR Policy

guusdk:


guusdk:
With https://github.com/dwd/xsf-web-docker/pull/1 in place, most of the changes (but not all) are no longer needed. I'll need to rework the PR.

Issue #321, opened 2017-06-02
Dead links : Sponsor


hbaltz:
Hi, 
  When you click on one of the sponsor logo in the foot page, you obtain an error 404 Not Found.
![xmpp_sponsor_link](https://cloud.githubusercontent.com/assets/15694673/26725474/fcfb5240-479e-11e7-8590-1d86a0a0d840.PNG)

Crispy33:
<a rel="license" href="http://creativecommons.org/licenses/by-sa/4.0/"><img alt="Creative Commons License" style="border-width:0" src="https://i.creativecommons.org/l/by-sa/4.0/88x31.png" /></a><br />This work is licensed under a <a rel="license" href="http://creativecommons.org/licenses/by-sa/4.0/">Creative Commons Attribution-ShareAlike 4.0 International License</a>.

guusdk:
Thanks for reporting this issue, @hbaltz !

@Crispy33 I do not get your comment. What are you trying to say?

guusdk:
This will likely be fixed by https://github.com/dwd/xsf-web-docker/pull/1

PR #327, opened 2017-06-14
Fix for #321 Dead Link: Sponsors

guusdk:


guusdk:
With https://github.com/dwd/xsf-web-docker/pull/1 in place, this PR should no longer be needed.

PR #329, opened 2017-06-28
Replacement for the dead link xmpp.net/directory.php

psy-q:


guusdk:
I don't know about this one. I somehow feel that we should have xmpp.net back, and link that instead... Thoughts?

horazont:
Having a dead link on the website is uncool in any case. Why not link to a usable list until we have xmpp.net back?

I’m for merging this and filing an issue which reminds us to change it back at some point.

(Let the bike-shed which list to use begin.)

guusdk:
Right you are. I had hoped to prevent this by having the Observatory restored by now. Sadly, that's not the case.

PR #340, opened 2017-08-03
software/libraries: Update gloox last_renewed

LNJ2:
I'm not the author of the library, but the last update is less than a year ago and I wanted to update this here. The date I've set is the latest release date (2017.02.26 / 1.0.20). Is this correct, or should it be the current date?

### DIFF
--- a/data/libraries.json
+++ b/data/libraries.json
@@ -140,7 +140,7 @@
         "url": "https://github.com/theozaurus/frabjous"
     },
     {
-        "last_renewed": null,
+        "last_renewed": "2017-08-03T10:19:16",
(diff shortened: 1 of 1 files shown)


guusdk:
The date reflects the date that the entry on this page was last updated / verified. You can thus use the current date (or use the script, as described in https://github.com/xsf/xmpp.org/blob/master/data/README.rst which will set the current date). You can use an older date, but that means that the entry has to be renewed sooner than in one year. And no, you can't set a future date. :)

LNJ2:
Ok, thank you. I'll change it to the current.

Issue #203, opened 2016-07-23
Various issues with the trademark pages


linkmauve:
These two pages, as linked from the main Jabber trademark page, don’t exist currently: https://xmpp.org/about/xsf/jabber-trademark/guidelines and https://xmpp.org/about/xsf/jabber-trademark/dispute-resolutions
The first one should be https://xmpp.org/about/xsf/jabber-trademark/usage-guidelines instead, I haven’t found the second one.

https://xmpp.org/about/xsf/jabber-trademark/pending-applications links to http://tracker.xmpp.org/browse/TRADEMARKS (which redirects to https://xmpp.org/), it should instead maybe list the open issues on github tagged with a trademark tag for example.

https://xmpp.org/about/xsf/jabber-trademark/who-needs-a-license says “described elsewhere on this web site”, this should be replaced with a link.

The various “please contact us” links point to https://xmpp.org/contact but it might be useful to add an id on the trademark paragraph to link directly to it.

The first link on https://xmpp.org/about/xsf/jabber-trademark/whats-required points to https://xmpp.org/about/xsf/jabber-trademark/trademark-license-agreement/ (with a final slash), which doesn’t exist, this one should redirect to the same URL without the final slash, or maybe just the link should be fixed.

ge0rg:
Please reopen the issue, we are not done yet. I'm quoting the still-open things from the first comment:

> https://xmpp.org/about/xsf/jabber-trademark/pending-applications links to http://tracker.xmpp.org/browse/TRADEMARKS (which redirects to https://xmpp.org/), it should instead maybe list the open issues on github tagged with a trademark tag for example.

Indeed, we need to improve the process and document it.

> https://xmpp.org/about/xsf/jabber-trademark/who-needs-a-license says “described elsewhere on this web site”, this should be replaced with a link.

I'm planning to re-read all the different documents regarding trademarks and to make a single point of entry to them.

> The various “please contact us” links point to https://xmpp.org/contact but it might be useful to add an id on the trademark paragraph to link directly to it.
> 
> The first link on https://xmpp.org/about/xsf/jabber-trademark/whats-required points to https://xmpp.org/about/xsf/jabber-trademark/trademark-license-agreement/ (with a final slash), which doesn’t exist, this one should redirect to the same URL without the final slash, or maybe just the link should be fixed.

Still needs fixing.

Echolon:
@ge0rg may you have time to review on this again?

ge0rg:
[...]

PR #342, opened 2017-08-03
Fix for #203 broken links

ge0rg:
This series moves the PDF files into /docs/ on the webroot and fixes the broken links called out in #203, as well as some others.

This does not change the actual content of the confusing licensing pages, this is subject to a future PR.


guusdk:
That last commit changes http://xmpp.org/about-xmpp/ into http://xmpp.org/about/ which appears to be a structural change. External references to the old URL will now fail (and, as there are blogposts touched by that commit, there likely are external links. Is this change needed?

ge0rg:
When I started with the change, I didn't realize that there is a permanent forward from `/about-xmpp/` to `/about/`, so technically, this change isn't needed except for SEO reasons. I tried not to touch anything infrastructural (i.e. actually move documents), just fixed the links to the new URL format.

guusdk:
Ah, that forward does not work in my Vagrant-based image, but it does work on the live website. With that, my reservation disappears.

PR #343, opened 2017-08-03
Please update Tigase as updated server software, we are currently in …

Dwisnewski:
…development of v7.2.0 and v7.1.1

Change line 203 from
"last_renewed": null,
to 
"last_renewed": 2017-08-03T12:26",

### DIFF
--- a/data/servers.json
+++ b/data/servers.json
@@ -200,7 +200,7 @@
         "url": "http://siemens-enterprise.com"
     },
     {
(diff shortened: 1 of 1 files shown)


guusdk:
I'd be happy to. Please know that having the software be under development is not a requirement to be listed. The only requirement is that the listing is updated once a year (which intends to purge the list of stale data)

PR #344, opened 2017-08-03
Update Tigase Clients as updated files

Dwisnewski:
Change "last-renewed" for Tigase clients to current date as well as add new versions of iOS and Android to clients list.

### DIFF
--- a/data/clients.json
+++ b/data/clients.json
@@ -504,15 +504,23 @@
         "url": "http://talkonaut.com"
     },
     {
-        "last_renewed": null,
-        "name": "Tigase Messenger",
+        "last_renewed": "2017-08-03T12:29:00",
+        "name": "Tigase Messenger for Android",
         "platforms": [
-            "Browser"
+            "Android",
         ],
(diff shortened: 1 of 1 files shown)


guusdk:
Continuous Integration fails to build this PR. I'm not exactly sure why. I've never seen a github diff use red letters. Did you use an incompatible character encoding?

Dwisnewski:
Nothing different that I'm aware of, I just used the field to push the replacement.  Let me check the push again.

Dwisnewski:
I found the errors, it was a problem with syntax did not close the "platforms" container properly which led to the issue, 2nd commit should fix now.

Dwisnewski:
Bad formatting errors, forgot that single lines in brackets do not need terminating commas.  Document should be good to go now.

guusdk:
Hmm, could you replace this with a single, link that refers to both platforms? I'd like to avoid having a list that lists some clients individually (per platform), while others do not. That'll lead to a tug of war, I'm afraid.

guusdk:
Gracias

PR #348, opened 2017-08-09
Docker image should be as small as possible.

guusdk:
This commit reduces the generated docker image roughly in half (to 300-something MB). It does so by:

- Replacing the base image (debian) by a smaller one (alpine);
- Introduce a .dockerignore file;
- Minimize layers by combining commands;

This was Linted via https://www.fromlatest.io/#/


.dockerignore
```diff
@@ -0,0 +1,12 @@
+*.swp
```
SamWhited: Probably `*.sw[op]` for maximum coverage (assuming this sports regex like things like gitignore)

Dockerfile
```diff
@@ -9,25 +9,29 @@
 
 # THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 #
+FROM gliderlabs/alpine:3.3
```
SamWhited: Who is gliderlabs? Should we use the official images instead (or are these official? I've been using just `alpine`)?

guusdk: I'm new to all this, and primarily done this using examples. From what I understand, gliderlab authors both, but the gliderlab one has more tooling?

guusdk:
I'm revoking this. The website is working as-is. The risk of breaking stuff is not worth the benefit of having this merged.

PR #353, opened 2017-08-16
Adding JC and Jonas to editor team.

guusdk:


guusdk:
As OK'd by Board in todays meeting.

PR #355, opened 2017-08-19
Update server list for AstraChat

jefryjacky:


horazont:
This looks like a hosted service and not like a freely available server software. Do we want to list those, too?

guusdk:
Perhaps we should create a new page, for hosted services?

horazont:
I was a bit in a hurry, so here are the issues I have with this PR:

* It advertises a hosted service, but this is a server *software* listing.
* The list of platforms most likely does not refer to the platforms the *server software* runs on, but the platforms the client software developed for the server runs on.

To be clear, I’m not set against advertising commercial software/services or hosted services in general (that’s not my thing to decide). Hosted services are simply, in my eyes, not fitting into the server *software* directory scheme.

I agree that a page for hosted services (where e.g. also the conversations.im thing could be listed) may make sense. I’ve heard throughout the net that there is demand for hosted services.

jefryjacky:
AstraChat server is exclusively sold on a licensed basis and Rockliffe does not host any AstraChat server accounts.

Here is the feature they provided in their server software http://www.astrachat.com/features.aspx

guusdk:
@jefryjacky are you associated with Rockliffe and/or AstraChat?

AstraChat server is advertised as a hosted or on-premise server solution - so it might actually fit both categories?

Can it federate with other XMPP servers, that are not AstraChat? If so, is there a publicly addressable XMPP domain hosted by AstraChat available?

jefryjacky:
Hi @guusdk ,

Yes I am associated with Rockliffe,

Rockliffe is not Hosting provider for XMPP software, so it is not hosted services category.
We only sell AstraChat server software and provide service to help the installation/integration for their customer.

Here is the link to purchase the AstraChat server software http://www.astrachat.com/purchase.aspx.

guusdk:
Can it federate with other XMPP servers, that are not AstraChat? If so, is there a publicly addressable XMPP domain hosted by AstraChat available?

jefryjacky:
Yes, the AstraChat server does support XMPP federation.  Our customers run their own AstraChat servers so I cannot provide a publicly addressable XMPP domain

horazont:
@jefryjacky Thanks for clearing this up.

I think this can be merged now.

jefryjacky:
you're welcome @horazont

PR #360, opened 2017-09-04
add xmpp.js library

sonnyp:
node-xmpp was rewritten and renamed xmpp.js


guusdk:
As the old domain name forwards to the new one, I'm inclined to accept that. :)

Thanks for the update!

PR #361, opened 2017-09-05
Prosody update

linkp:
Updated platforms. Windows is no longer supported, and all 3 main BSD releases (Free/Net/Open) have Prosody in their packages.


guusdk:
Thank you!

PR #363, opened 2017-09-07
Fixing the XMPP logo

guusdk:
Where the two arms of the logo cross, everything from the right-hand arm crosses under the left arm, with the exception of the darkest orange bit. That bit passes over the left arm. 

Some argue that this might be intentional. As the blue colored bit on the right arm again passes under, I feel that this is a bug (if the blue had _also_ gone over, there would be an argument that the left hand arm goes _through_ the right hand arm - but it doesn't now).

This commit modifies the logo in such a way that all of the right hand arm goes under the left hand arm.

logo-after-before

And as I'm crap at embedding SVG in Github Markdown, I've created this screenshot (at 200% zoom level). Left is my proposed fix, right is the original.


Kev:
I'm not convinced that it was broken to start with :)

guusdk:
It is. The current logo is inconsistent. Either the blue part on the right should also be raised (creating an effect of one arm 'running through' the other), or the entire right half of the logo should be lowered (as I've done in this PR).

ge0rg:
Here is an amateurish attempt at rendering the third variant:

xmpp-logo svg

guusdk:
Thanks @ge0rg !  That's the other option that I referred to

I don't like this one much. It's a bit Escher-esque: the bottom half now looks three dimensional, while the top half of the logo does not.

Although I prefer this over the unmodified one, I like the one that I created in this PR best.

ge0rg:
I've come to the realization that I don't like any of the three. The two arms just don't look properly connected in any one of them.

ThurahT:
I think Guus suggestion is great. I interpret the logo as two horns tooting/communicating in different directions and should not be melted or half merged into each other.

jcbrand:
Here's a modified version where the blue line also goes over:

xmpp

mathieui:
I'm not sure the current logo is an accident, but on the other hand I do prefer the updates shown in this PR (both the one from @guusdk or the one from @ge0rg are better and more consistent than what we currently have). I do think the one from Guus is cleaner and simpler. It would be really nice if we had the opinion of the original designer, if they can still be found.

guusdk:
I've reached out to the original designer. I'll post here when I get his feedback.

guusdk:
The feedback of the designer was this: he confirms that it (the orange crossover) appears intentional, but thinks it looks odd and unbalanced. He cannot recall where it came from, as it all happened long ago. He did send over various drafts of the design, where the orange bit appears only in the last one.

Sevefdp:
Hey guys, I know I'm a bit late to the party, but it is because I've been trying to find something.
When I saw this Pull Request, I realized how the logo looked like. In my mind, the logo was actually like @guusdk is suggesting here. And that's because I have seen the logo with the two arms on top of each other. 
I could find an example, like this one: https://xmpp.org/images/promo/xmpp_server_guide_2017.pdf

This, arises some questions in me: 
Does the author of that guide modified the logo for that purpose? Or are we using the wrong version of the logo?

Anyway, regarding my personal opinion on the logo, I'm all for the new change.

horazont:
I’m all in favour of Guus’ new version.

Coincidental, I have dug around the ML archives from really back then, and given that the process in which the logo was chosen created quite some dismay around members (because they didn’t feel included in the decision), we should consider deciding the change in an all-member meeting.

stpeter:
People are always unhappy. I recall that we mostly delegated this decision to Matt (who had some design sense, unlike everyone else). This is a minor adjustment that doesn't justify a full-XSF meeting or decision. But feel free to raise it on the members@ list....

guusdk:
I've just sent out a notification on members@

penguineer:
Thanks for the mail on members@, which brought me to this discussion.

+1 for the PR, the logo looks much clearer and given the fact that it is usually used in a much smaller scale, IMHO we should not have such tiny features. 

(Frankly, in 10 years of using XMPP I never realized  that the arcs do not just overlap. Or maybe some third-party appearances of the logo have been corrected, too.)

sonnyp:
+1 for the PR 

I've wondered in the past why there was an unnatural overlap in the logo.

agnauck:
It's intended. When we worked on the logo (long time ago) the goal was to impress connectivity.
Connect people, devices, services etc...

guusdk:
Even if the original bit is intended - does anyone have strong feelings against the improvement in this PR?

@agnauck, you're referring to "we" - suggesting that others where involved. Those people are most likely to disagree with this PR. Could you reach out to them, ask them for feedback?

Kev:
> Even if the original bit is intended - does anyone have strong feelings against the improvement in this PR?

Me, for what it's worth. I think just overlaying the ribbons looks bland.

agnauck:
@guusdk the logo creation was initiated by the board in the 2006/2007 term.

Discussions in the memberlist start in July 2007 here:
https://mail.jabber.org/pipermail/members/2007-July/thread.html

penguineer:
> I think just overlaying the ribbons looks bland.

On the other hand, the fancy overlap looks like a scaling/display error when the logo isn't shown in full size.

Independent from the original intent a simplification seems to be within logo design trends.

horazont:
As I expected, there are voices against this change.

Given that there will be full member meetings soon anyways due to board / council elections, maybe it makes sense to simply have all members vote?

That’s fair, and given that there will be such meetings rather soon-ish, it doesn’t seem (to me, @agnauck as organizer may object) too much hassle to me.

I also strongly feel that github is not the right venue to discuss this (members@ seems appropriate), just like the xeps repository is not the right place for technical discussion (we have standards@ for that).

guusdk:
@horazont: last week, I've sent an email to member@ inviting people to discuss the subject here. That as a result, discussion is taking place here, is hardly surprising. ;) I'm not sure if the members list is a _better_ venue to discuss these things, but I'd be happy to do that there in the future, if that's preferred.

With 60+ people, consensus is guaranteed to be absent in pretty much any decision. Topics that relate to style are quite impossible in that respect. We shouldn't aim for consensus here, I think. We can, however, make sure that everyone can voice their opinion, and weigh those.

penguineer:
Maybe we can get a rough consensus (RFC7282).

guusdk:
As suggested by @Kev, I've asked Board to decide on this. It's been scheduled here: https://trello.com/c/KMFd33fZ/285-approve-guuss-logo-amendments

guusdk:
Board discussed and voted on this this in [yesterdays meeting](https://mail.jabber.org/pipermail/standards/2017-September/033409.html).

Votes were: +1, +1, 0. Ralph would like to see votes from the absent board members (@Nyco, @arcriley)

arcriley:
Enthusiastic +1

Nyco:
And @Nyco has finally read all those comments!
As a result, I believe as well that this PR is an improvement.
So that's a +1 for me.
And a big thanks to @guusdk ! ;-)

PR #365, opened 2017-09-13
Adding all work teams.

guusdk:

PR #367, opened 2017-09-16
Source member lists from a single JSON file

### LINKED ISSUE #366: Create reference list of XSF members.
The website keeps a record of XSF members, with workteam affiliation in various places:

- the members list
- the pages of each individual team

Updates tend to modify one, but not the other. It would be good if we could have one list, and generate the various content from that. Somewhat similar to what Jonas did for the software listing.

horazont:
In the same spirit as #269, this PR allows generation of membership lists from a single JSON source, located at ``data/members.json``.

It is applied to the global memberlist (/about/xsf/members.html) as well as team pages (editor and infrastructure).

Fixes #366.


guusdk:
Awesome! That'll do wonders for keeping a consistent record of our members. Thanks a lot!

I've just merged https://github.com/xsf/xmpp.org/pull/365, which adds the SCAM team page, but also introduces a conflict with this PR (three people got a new role, that was previously non-existent).

My Python knowledge is non-existing, but I don't believe you apply any ordering. If it's not to hard, it'd be nice to order all generated lists on family name.

The Council and Board members are listed on [About > XMPP Standards Foundation (XSF)](https://github.com/xsf/xmpp.org/blob/master/content/pages/about/xmpp-standards-foundation.md). Perhaps we could use templates there too? These pages also have a short biography for each member. That could be added to the json data source, perhaps.

guusdk:
@agnauck, as the keeper of data in this regard: do you have any suggestions?

horazont:
I deliberately decided to not apply any ordering. Detecting which part of the name is the to-be-ordered-by family name is not trivial ([see also](http://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/)). I can add some heuristic and an additional attribute for the member records to override it in case of need.

Regarding bios for council & board, that’ll be a bit trickier, especially since it contains markup. We should consider whether it’s worth the trouble.

I will rebase on #365 soon.

guusdk:
I was thinking of adding an optional 'order-by' field, and if absent, order by the last word after splitting on whitespace - but you're right - this might be more trouble than fixing things. Same goes for the biographies.

horazont:
@guusdk Did things!

The team role names in the member list are now links, also the SCAM team page was ported to the new template.

guusdk:
Hmm, the SCAM page displays `$MEMBERS_LIST$` instead of the member list.

horazont:
Right, that was a typo. Good catch.

horazont:
Tested with docker, editor, infrastructure and SCAM team pages work as intended. Member list looks ok.

guusdk:
Thank you!

agnauck:
agree that ordering by last name is not trivial. Its fine as is, and can take care of the ordering in the JSON

PR #368, opened 2017-09-19
Add go-xmpp library

petermbenjamin:


SamWhited:
From the docs:

> Package xmpp implements a simple Google Talk client using the XMPP protocol described in RFC 3920 and RFC 3921. 

Do we want to list something that implements an old version of XMPP and is specifically targeted at Google Talk with all its idiosyncrasies?

petermbenjamin:
Oh, I apologize. I may have committed an incorrect URL. I was evaluating a few different XMPP Go libraries and I intended to commit: https://godoc.org/google.golang.org/appengine/xmpp

Is this more appropriate?

SamWhited:
If I'm not mistaken that one isn't a general purpose XMPP library and does some things against the spec (such as allowing random XML in the body of a message).

That being said, I'm not sure that either of these are necessarily inappropriate or who actually makes that decision, it's just worth noting that there are some caveats. I'm not aware of a good general purpose / modern XMPP library in Go, but if you find one I'd love to know about it.

guusdk:
I don't think there are any formal acceptance criteria. As for who makes the decision to add it to this list: that appears to be the person who first hits either the "Merge" or "Close" button in GitHub. That appears to be mostly me these days (although I'd _love_ some others to join in here). 

Personally, I'm primarily looking if the suggested addition is a bit of software that can be reused by others (in other words, isn't used within a silo). Google Talk libraries, for instance, would pretty much fall into that category these days. Other criteria have simply not come up yet. In perspective of this PR: some level of feature-completeness appears to be desirable.

As a side-note: this PR caused a CI failure:

> ERROR: entry 'go-xmpp': should be placed behind 'Z-XMPP' (all entries must be ordered alphabetically by case-folded name)

Although you can manually edit the source (if you're willing to do a couple of CI iterations), we also provide a script that will generate a valid change in the first go.

horazont:
@petermbenjamin 

Can you please: 

1. clarify whether the library you propose is general-purpose and can be used outside of silos?
2. confirm whether the library encourages violation of standards, as claimed by @SamWhited?
3. fix the CI failure @guusdk mentioned

Thanks!

petermbenjamin:
@horazont - I prefer reading enumerated lists over paragraphs anyway 😄

I have closed the pull-request because of concerns -- that I agree with -- raised by members who know more about xmpp than I do.

1. Neither library I proposed seemed to be general-purpose enough
1. Not familiar enough the XMPP spec to be able to determine this in an efficient time manner. I would have to spend more time than I can afford right now to determine if the library encourages violation of standards. So, I take @SamWhited's word for it 🙂
1. No need to fix CI since I'm closing the PR.

Thank you.

SamWhited:
Thanks @petermbenjamin! If you find a general purpose XMPP library that's relatively complete be sure to let us know! I made some changes to `encoding/xml` for Go 1.10 that should make implementing streaming protocols like XMPP easier, so hopefully we'll see some in the future.

PR #369, opened 2017-09-23
Replace http://mail.jabber.org with https://mail.jabber.org

MilanKral:
mail.jabber.org supports HTTPS.
Please replace all HTTP links with HTTPS


guusdk:
Thank you!

PR #370, opened 2017-09-28
Replace favicons

guusdk:
This commit replaces the favicons used on the website. Unlike the old favicons, the new ones are
centered, and have a transparent background.

This commit replaces, but also removes, a lot of the older favicons. The new package is much
lighter, yet keeping the same compatibility level as before.

The package was generated using the SVG logo, on https://realfavicongenerator.net


guusdk:
@lloydwatkin I think you touched these last. Does this look okay to you?

lloydwatkin:
I'd agree looks fine to me. I think last time I just took the existing favicon and run it through a site to generate all the crazy numbers of apple variants (if only there were standards!).

PR #371, opened 2017-09-29
Move favicons to content root

guusdk:
Most clients expect the various favicons to exist in the content root. Instead of maintaining redirection
scripts, simply placing them in the content root does the trick too.


guusdk:
(that didn't work as expected)

PR #379, opened 2017-10-13
Added CoyIM to list

Milestacular:


horazont:
Hey,

Thanks for your submission.

Could you please fix a few things?

1. This is invalid JSON, it lacks a "," before your new entry.
2. The "last_renewed" field must not be ``null`` for your listing to be effective. Set it to the current date and time like other entries do.
3. Move the entry to the correct location in the list (the list is in alphabetical order, so it should be before "Dino").

Thanks!

guusdk:
This PR seems to be dead. As it also needs a rebase now, I'm closing it. Please create a new one when the intended change is still desired.

PR #383, opened 2017-11-13
add MatriX vNext library

agnauck:
add MatriX vNext library to libraries.json

### DIFF
--- a/data/libraries.json
+++ b/data/libraries.json
@@ -325,6 +325,16 @@
         "url": "https://github.com/agnauck/matrix-js"
     },
     {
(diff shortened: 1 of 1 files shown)


guusdk:
Currently, these are listed:

- MatriX | .net / C# / Mono
- MatriX JS | JavaScript / TypeScript

Does MatriX vNext replace MatriX, or are they distinct projects?

agnauck:
They are distinct projects. MatriX vNext is the "successor" or MatriX dedicated to netCore and netStandard.
While netCore is very new it may take years until everyone is able to update. So they will side by side for a while and will both be fully supported.

PR #387, opened 2017-12-09
update Q4-2017

agnauck:
update memberlist after Q4-2017 voting process

### DIFF
--- a/data/members.json
+++ b/data/members.json
@@ -67,11 +67,7 @@
             "name": "Dave Cridland",
             "roles": ["council"]
         },
-
-        {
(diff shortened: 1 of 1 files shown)


horazont:
@agnauck AFAIK the list isn’t sorted at any later place, so the order in the JSON will be the order on the website. Is it intended that Maxime is at the end of the list?

agnauck:
will fix then. My assumption was that some basic sorting is alöready implemented ;-)

horazont:
Thanks! We decided against sorting in the code because sorting names isn’t exactly trivial.

PR #388, opened 2017-12-22
2018, 22nd summit

Nyco:


horazont:
Syntactically okay, is it sensible to have a future event under history though?

Nyco:
Ah, right ;-) We'll keep for the 5th of Feb then?

More seriously, I was wondering what other impactful events happened this year. I guess GSoC, plus more I presume. I doubt we should mention the website+backup failures ;-)

PR #393, opened 2018-01-05
Update clients.json

TheSZ11:
ChatSecure does not support Android anymore, amended list to reflect that.

More info: https://chatsecure.org/faq/


guusdk:
Thanks!

PR #396, opened 2018-01-18
Added an entry for UWPX

COM8:
Added the UWPX XMPP client.

### DIFF
--- a/data/clients.json
+++ b/data/clients.json
@@ -615,5 +615,11 @@
             "iOS"
         ],
         "url": "https://zom.im/"
+    },
+    {
+        "last_renewed": null,
+        "platforms": ["Windows 10 (UWP)"],
+        "name": "UWPX",
+        "url": "uwpx.org"
     }
 ]
--- a/data/clients.json
+++ b/data/clients.json
@@ -617,9 +617,9 @@
         "url": "https://zom.im/"
     },
     {
-        "last_renewed": null,
-        "platforms": ["Windows 10 (UWP)"],
+        "last_renewed": "2018-01-18T08:37:00",
         "name": "UWPX",
+        "platforms": ["Windows 10 (UWP)"],
         "url": "uwpx.org"
     }
 ]


guusdk:
Hey, thanks for your contribution. I think you've manually modified the json file - which is fine, but often leads to errors that will cause our parser to fail. Please address the errors as reported by Travis:

> ERROR: entry 'UWPX': should be placed behind 'Zom' (all entries must be ordered alphabetically by case-folded name)
> ERROR: entry 'UWPX': undefined platforms: 'Windows 10 (UWP)' (the allowed platforms are listed in platforms.json. If you think a platform is missing add it and mention it in your Pull Request)
> Found 2 severe violations. Please fix them.

Also, its website lists the client as an early Alpha. I'm reluctant to add non-feature complete clients to our list, as that's likely to disappoint users. Perhaps we should add this client only after the first proper release?

horazont:
(note that the error message is unclear and UWPX should be placed **in front of** Zom, not behind.)

COM8:
Ok, thanks I will creata a new pull request once UWPX reached the stable phase.
 > (note that the error message is unclear and UWPX should be placed in front of Zom, not behind.)

If you follow the [documentation](https://github.com/xsf/xmpp.org/blob/master/data/README.rst) it says:
> [...] Insert it into the top-level JSON Array as **last** element by [...]

😉

horazont:
@COM8 The same documentation also says (boldface mine):

> Insert it into the top-level JSON Array as last element by adding a comma after the last , and then pasting the above template with your modifications. **Use the tool as described** in the previous section to perform a renewal (**this will sort the list correctly** to minimize future diffs).

And in the next paragraph (boldface original, emphasis mine):

> **If you do not use the tool**, make sure that you set the ``last_renewed`` key manually to the current date (as seen in other entries) in UTC and _adhere to the sorting requirements of the JSON file_. You can use the ``lint-list.py`` tool to verify that everything is in order. If ``lint-list.py`` complains, the Travis CI will reject your Pull Request.

😉

Issue #399, opened 2018-01-30
Psi/Psi+ are missing on https://xmpp.org/software/clients.html


Ri0n:

horazont:
Please see [the new readme](https://github.com/xsf/xmpp.org/tree/master/data#software-directory-readme).

If you’re affiliated in some way with Psi/Psi+, feel free to open a PR.

PR #400, opened 2018-01-30
Update Metronome IM, data/servers.json entry.

maranda:
Also xmpp.org servers list isn't using the entries from this file.

### DIFF
--- a/data/servers.json
+++ b/data/servers.json
@@ -144,13 +144,13 @@
         "url": "http://kwickserver.info"
     },
     {
-        "last_renewed": null,
+        "last_renewed": "2018-01-30T07:27:16",
         "name": "Metronome IM",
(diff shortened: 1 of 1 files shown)


horazont:
@maranda I don’t understand your claim that the server list doesn’t use the entries (I’m pretty confident it does). Why do you think so?

maranda:
@horazont [does it?](https://xmpp.org/software/servers.html)

maranda:
@horazont it doesn't look to reflect the contents, but it could be an optical illusion :)

horazont:
@maranda [Seems to me like it.](https://xmpp.org/2017/03/new-xmpp-software-listing-rules/)

horazont:
@maranda Maybe also see #399.

PR #401, opened 2018-01-30
Renewed Psi/Psi+ clients

Ri0n:
Just in case about me (I don't if next links are publicly available. I don't have other github accounts to check)
https://github.com/orgs/psi-im/people/Ri0n
https://github.com/orgs/psi-plus/people/Ri0n

### DIFF
--- a/data/clients.json
+++ b/data/clients.json
@@ -401,24 +401,24 @@
         "url": "http://profanity.im"
     },
     {
-        "last_renewed": null,
+        "last_renewed": "2018-01-30T09:21:32",
         "name": "Psi",
         "platforms": [
             "Linux",
             "macOS",
             "Windows"
         ],
-        "url": "http://psi-im.org"
+        "url": "https://psi-im.org"
     },
     {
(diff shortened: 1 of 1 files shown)


guusdk:
Thanks for your contribution! I'm not sure what you'd like us to do with the account names that you provided though.

Ri0n:
well, I thought only a project's owner can make such a pull request. and I'm the current owner of Psi and Psi+.

guusdk:
Some people might be more strict in accepting PRs than I am. ;)

horazont:
FWIW, we don’t enforce the project owner rule (if there even is one). At least I don’t. I like to ask why they add a client if they’re not the owner, but usually I don’t have to because the client has enough info on the website to make sense.

Kev:
The reason for that rule was mostly to stop one person going around and submitting a tonne of things unbeknownst to the owners.( "I think the list should have all these things on it, and I'm adding them all" type people). It doesn't really need strict enforcing, but that's the rationale.

PR #402, opened 2018-01-30
yaxim + Bruno renewal, fix Jabber™ trademark

ge0rg:


guusdk:
See, that was fast&easy! ;)

PR #409, opened 2018-02-28
added CC By-Sa license to the newsletter

goffi-contrib:
This will allow to copy the newsletter and translate it on other
websites.

I have asked commteam@ and specifically to @jcbrand  if it was OK.

### DIFF
--- a/content/posts/newsletter/2018-03.md
+++ b/content/posts/newsletter/2018-03.md
@@ -75,3 +75,4 @@ There were [3.3.x releases of Converse.js](https://opkode.com/blog/converse-3.3.
 
 Fork Awesome, a fork of Font Awesome which includes more community contributed icons, now has an [XMPP icon](https://forkawesome.github.io/Fork-Awesome/icon/xmpp/?utm_source=xmppnewsletter).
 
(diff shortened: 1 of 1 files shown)


jcbrand:
Given that the whole site was originally licensed under CC and that the CC license will again be added site-wide, I suppose this isn't necessary anymore, right?

goffi-contrib:
@jcbrand indeed, this P.R. is probably not needed anymore, but for now the CC 2.5 mention is not on the website, so we still can't use the newsletter on other websites.

guusdk:
Board has discussed this in last weeks meeting. Board agrees to have the newsletter published under the Attribution-ShareAlike 4.0 International (CC BY-SA 4.0) license.

Echolon:
All the recent Newsletters were published under 4.0, I think this can be closed.

Echolon:
I will close this as we do so as said. Please come back if something else comes up or is treated not correctly in your view.

See for example: https://github.com/xsf/xmpp.org/blob/newsletter-july2020/content/posts/newsletter/2020-07-01.md

PR #416, opened 2018-03-20
Added jackal golang server to servers.json

ortuman:


horazont:
Sorry for the delay. On an unrelated note, you may want to revisit stream compression and maybe drop it, due to the security implications.

PR #425, opened 2018-03-31
Update clients.json

darkpsy3934:
Updated Pidgin as it is still a valid XMPP client

### DIFF
--- a/data/clients.json
+++ b/data/clients.json
@@ -372,7 +372,7 @@
         "url": "http://oneteam.im"
     },
     {
-        "last_renewed": null,
+        "last_renewed": "2018-03-31T06:00:00",
         "name": "Pidgin",
         "platforms": [
             "Linux",


ge0rg:
While it is true that Pidgin has come back from the dead, its XMPP support is so bad that I consider it actively harmful to the XMPP ecosystem. [Carbons](https://xmpp.org/extensions/xep-0280.html) are **essential** if you have more than one client, and Pidgin fails to support them for [over five years now](https://developer.pidgin.im/ticket/15508). And I don't even want to get started about its sudden popup windows stealing focus from whatever you were just doing.

While this is a clear violation of the XSF's neutrality stance, **I am strongly against taking Pidgin onto the list.**

guusdk:
I'm limping on two thoughts here: @ge0rg's point is obvious, but on the other hand, Pidgin is a popular client. Adding it to our list can attract more potential users (that might add voices to the ongoing effort of getting their XMPP implementation up-to-date). Maybe we can add some kind of "interop compliance" indicator?

horazont:
@darkpsy3934 We do want projects to renew their listing themselves. I normally do a rough Github based check to find that out, but that fails with you. Can you provide some evidence that you’re involved with Pidgin in any way?

Thank you for your help.

PR #426, opened 2018-04-05
Added League of Legends to projects list

kawaii:
Was suggested to make this PR by Ge0rge in the Prosody developers MUC. Always good to raise awareness of how widely used XMPP actually is. :slightly_smiling_face:


ge0rg:
Thanks very much! My comment was a bit tongue in cheek, but now you made me follow up :)

PR #443, opened 2018-06-05
Getting Started: clean-up

ge0rg:
* make links https
* remove Adium (which is not on the client list) and Pidgin (because it still sucks)
* add Monal and Converse.js
* add yax.im as a public server


guusdk:
Karma has been restored! :)

PR #446, opened 2018-06-06
Add Kaidan to clients list

LNJ2:
Theoretically Kaidan also compiles for macOS, Windows and iOS, but these are not tested. Also, Kaidan is running on Plasma Mobile - Is there a need for adding smth. as `Linux Mobile` / `Plasma Mobile` as platform?

### DIFF
--- a/data/clients.json
+++ b/data/clients.json
@@ -297,8 +297,8 @@
         "last_renewed": "2018-06-06T20:15:00",
         "name": "Kaidan",
         "platforms": [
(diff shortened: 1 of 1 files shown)


guusdk:
Thanks for your addition. I've merged it as-is. We've opted on previous occasions to not name the more exotic platforms on this page. They clutter up the page a lot, but offer little additional value. Besides, the users of those platforms are likely aware of their platform-specific software.

PR #447, opened 2018-06-10
Extended functionality to display the support of some XEPs in the clients table.

zak333:
I added two new fields 'features' and 'comments' for every client in clients.json to be able to add a list of XEPs (or something else) which this client supports. The comments dict is optional and may contain extra text for some feature, like 'partial support'.

For managing possible features, a new file features.json contains all possible features, their name, url and an 'order' parameter. The latter is used to decide in which order this feature should be displayed in the table. order=0 means the feature is not displayed. features.json does not contain all XEPs yet, but this can be extended as needed of course.

For now I tried to fill the clients.json with features for AstraChat, Bruno, Dino, Gajim, Pidgin and yaxim as an example. If a list of features is given for a client, these features get a tick icon in the table, all others get a red cross icon. Comments are given as footnotes below the table and in the HTML title tag of the icon.
If there is no feature list set for a client, (i.e. features = null) there will be question marks displayed for every feature.

I updated the update-entry.py as well to include a --set-feature parameter. A comment may be given by adding it with a colon afterwards (e.g. --set-feature "XEP-0123:using plugin").

I'd be happy to get some feedback on this, possibly there can be some fine tuning on the layout, and of course there is only space to display some features/XEPs as an example, so we can discuss what columns we want to have. The idea is to give an overview on what clients have support for modern XMPP features.

It should be easily possible to add this for the servers support list later as well.

Edit: I had to add entries for 'features' and 'comments' to servers.json and libraries.json as well to satisfy the lint-list.py tool.

Edit2: See screenshot here:

screenshot


horazont:
This looks like great work!

I think we’ll indeed need a discussion on what we want to show in that table. Listing XEPs is probably not a good thing to do, because they’re rather technical.

Instead, we should list features (without XEP numbers), following common terminology. My suggestions:

* Group chats (= XEP-0045)
* End-to-end-encryption / OMEMO (= XEP-0384)
* Avatars (= XEP-0084 + XEP-0153)
* Multi-Device Support (= Carbons + MAM)
* Mobile-friendly (= Stream Management + Client State Indication)

guusdk:
Thank you for putting in the effort! I am, however, having concerns of having this feature. We've discussed doing this before (on multiple occasions). We ended up not doing this, for mostly these two reasons:

- It will be a daunting task to keep the provided information complete and accurate.
- In the past, listing information like this 'motivated' implementors to do the bare minimum to have the box ticked, but only for the sake of ticking this box. That leads to a misrepresentation of the implementation 'quality' in our list, but also appeared to trigger some kind of war-of-the-marketing-departments.

I'm not sure if we should have this list on our website. Perhaps a better place for this is the automatic tooling, such as the one that's being created as part of @RishiRaj22's GSoC project at https://github.com/iNPUTmice/caas

zak333:
> It will be a daunting task to keep the provided information complete and accurate.

You are right, it is probably not possible to keep all the information perfectly correct, but I still think the information gain is more important for users than the possibility of having some inaccurate tick somewhere. If someone recognizes an error later, it can easily be corrected.
A note "This information is supplied without liability." should clear this up.

> In the past, listing information like this 'motivated' implementors to do the bare minimum to have the box ticked, but only for the sake of ticking this box. That leads to a misrepresentation of the implementation 'quality' in our list, but also appeared to trigger some kind of war-of-the-marketing-departments.

Isn't this a positive effect? The XEPs are clearly defined (at least mostly I hope), so what's wrong with having a "bare minimum" supported XEP in a client in contrast to no support at all?

I don't understand how a list with clients where some have no support for modern XMPP features is of better 'quality', if this bad support is not evident for users.

horazont:
> > In the past, listing information like this 'motivated' implementors to do the bare minimum to have the box ticked, but only for the sake of ticking this box. That leads to a misrepresentation of the implementation 'quality' in our list, but also appeared to trigger some kind of war-of-the-marketing-departments.
> 
> Isn't this a positive effect? The XEPs are clearly defined (at least mostly I hope), so what's wrong with having a "bare minimum" supported XEP in a client in contrast to no support at all?

It depends. The [server compliance tester](https://conversations.im/compliance/) thing from @iNPUTmice for example has also had negative or at least not-positive effects. Those were of two types:

- Operators who simply don’t fully understand what they are doing (I feel that this has happened a lot with [XEP-0368](https://xmpp.org/extensions/xep-0368.html)) and want to tick the box.
- Operators who just load a module which only advertises a feature, but which needs additional things to do something useful (e.g. [XEP-0352](https://xmpp.org/extensions/xep-0352.html)). Since in those types of specs everything is implementation-defined (server-side at least), a tickmark that the spec itself has been implemented doesn’t really tell you a lot.

zak333:
Okay I see the problem now. But I still think the general goal to publish XMPP features of clients should be pursued. Otherwise we could as well decide to not publish any client at all because people chosing a poor client might get a wrong impression of what XMPP can do today which is a negative effect for sure.

I like the approach of @horazont to not even mention any XEP directly but give a descriptive text of what a client can do (or not do). Maybe we can reduce the negative impact to developers then in some way. We can further concentrate on features that have a clear "can-do" or "cannot-do" indication (i.e. leave out XEP-0368 and XEP-0352).

Ppjet6:
It seems to me that this is going to be superseeded by DOAP/#594 in most ways? It doesn't solve most of the issues mentioned here, but it makes gathering data a bit easier on the XSF as it's now the project's responsibility.

Should this be discussed somewhat? so that we don't leave PRs lingering around for years :)

zak333:
closing due to little interest

PR #453, opened 2018-06-30
Refresh and add the Movim client to the website

edhelas:
Refresh the Movim last_renewed datetime
Add Movim in the list of Clients supporting MUC, Pubsub and Jingle


guusdk:
Thanks!

PR #454, opened 2018-07-05
Add 404.city Supports all useful XEPs

5269:
Add 404.city . One of the forward public servers on Ejabberd. Supports all useful XEPs. The server with open registrations in XMPP-clients.
XEP-0163: PEP avatars and OMEMO
XEP-0191: Blocking Command
XEP-0198: Stream Management
XEP-0237:Roster Versioning
XEP-0280: Message Carbons
XEP-0313: MAM
XEP-0352: Client State Indication
XEP-0357: Push
XEP-0363: HTTP file upload

### DIFF
--- a/content/pages/getting-started/_index.md
+++ b/content/pages/getting-started/_index.md
@@ -31,6 +31,7 @@ As with e-mail, you'll need an account with a service provider. Again, there are
 * [JWChat](https://accounts.jwchat.org/)
 * [jabber.at](https://jabber.at/account/register/)
 * [yax.im](https://yaxim.org/yax.im/)
+* [404.city](https://404.city/)
 
 # 3. Log in! That's it!
 


guusdk:
Thanks for your contribution. After some discussion in the XSF muc, I think we should pass on this one. Some of the discussions are logged at http://logs.xmpp.org/xsf/2018-07-05/#14:03:32

horazont:
As a follow-up, the original open server list where anyone can be listed is at [xmpp.net](https://xmpp.net).

I would much prefer if we can get that list to be usable again, because this also allows to show information about servers (such as certificate status, and possibly in the future XEP support).

However, the listing and the service itself is still not reliable. If anyone would want step up and help with the maintenance of that service, that’d be great and helpful and settle this debate.

404city:
Yes, I agree with that. But unfortunately the XMPP.net server is unstable and new servers cannot be added there. compliance.conversations.im is stable and anyone can add their server there.
I suggest adding compliance.conversations.im until xmpp.net works well.
The list of servers list.jabber.at does not allow to add new servers freely

PR #458, opened 2018-08-29
Smack is used in Openfire, Strophe.js is used in Converse

marclaporte:


guusdk:
Thanks Marc!

guusdk:
(Also, Smack isn't used in Openfire, but hey ;-) )

marclaporte:
oups, right. Conversations I should have said :-)

guusdk:
(also not correct ;) )

marclaporte:
ok, I will say nothing more :)

PR #459, opened 2018-08-29
Extending Openfire. I use daily.

marclaporte:


guusdk:
Open...fire you say? Sounds promising. I should look into that. ;)

PR #460, opened 2018-08-30
Some XMPP client updates

marclaporte:
Thunderbird: https://support.mozilla.org/en-US/kb/instant-messaging-and-chat
Spark: I tried recently. It works. Not a ton of development, but it's chugging along.
Jitsi: I use daily. Still one of the better XMPP clients even if little recent development. Now called Jitsi Desktop to avoid confusion with Jitsi Meet
Converse.js now prefers to be called Converse. Version 3 is active, and v4 is coming soon.
MirandaNG just had a release: https://www.miranda-ng.org/en/news/0-95-8-1

### DIFF
--- a/data/clients.json
+++ b/data/clients.json
@@ -122,8 +122,8 @@
         "url": "https://github.com/siacs/Conversations"
     },
     {
-        "last_renewed": "2017-04-19T14:37:00",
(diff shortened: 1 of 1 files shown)


guusdk:
Thanks Marc!

PR #466, opened 2018-09-23
servers.json: remove MacOS from supported platforms.

maranda:
I don't know of any successful MacOS deployment ever neither I did try myself.


guusdk:
Thanks!

PR #471, opened 2018-10-03
Updating myths.md to point to correct URL

atcwells:
URL has apparently changed from:

http://balisage.net/Proceedings/vol10/html/Lee01/BalisageVol10-Lee01.html

to:

https://www.balisage.net/Proceedings/vol10/html/Lee01/BalisageVol10-Lee01.html


guusdk:
Thank you!

PR #476, opened 2018-11-06
Create a list for languages the newsletter is translated to

SeveFP:
How it looks:
screenshot_20181106_190637


jcbrand:
@SeveFP Thanks, but this makes it look like French is the only available language.

Perhaps rename the heading to `Available Translations` or something similar to make it clear that the French version is in addition to the English version.

SeveFP:
@jcbrand you are completely right. It didn't feel 100% correct to me, but I couldn't figure out what was the problem.

I'm sending a change right away. Thank you! :hugs:

PR #481, opened 2018-11-18
clients: Add new platforms to Kaidan

LNJ2:
Kaidan now also supports iOS, macOS, Ubuntu Touch and Windows. There haven't been
released official builds yet though (they will come with the next release).

### DIFF
--- a/data/clients.json
+++ b/data/clients.json
@@ -294,13 +294,17 @@
         "url": "http://kadu.net"
     },
     {
-        "last_renewed": "2018-06-06T20:15:00",
+        "last_renewed": "2018-12-04T15:00:00",
         "name": "Kaidan",
         "platforms": [
             "Android",
-            "Linux"
+            "iOS",
+            "Linux",
+            "macOS",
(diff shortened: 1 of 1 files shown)


LNJ2:
I added Ubuntu Touch as platform, because I don't think it's still "normal" Linux (Desktop). Also it has its own packaging format and [store](https://open-store.io/).

LNJ2:
If you're unsure about Ubuntu Touch, I can also remove it for now.

guusdk:
Ah, I forgot about this one, sorry for the delay.

Indeed - we generally don't add more platforms that we already have. The rationale behind that is that otherwise, we'd have an ever-growing list of platforms, of which most are pretty obscure. People that are using such a niche platform are likely competent enough in finding software that fits their platform.

Maybe you want to add 'other' ?

LNJ2:
Other seems to be a good idea :)

LNJ2:
Updated

guusdk:
Thank you!

PR #358, opened 2017-08-29
Apply for trademark sublicense for JabberCat

horazont:
The personal details have been supplied to trademark@jabber.org, hopefully reaching @stpeter as intended. If not, please let me know via email (jonas@wielicki.name) or other means.

(See also #200, if you’re wondering on why this PR was made and how that fits into the trademark process.)

PR #495, opened 2019-01-03
Replaced defunct tracker for Trademark applications.

guusdk:
The tracker that is linked to in the old version appears to be gone.

### DIFF
--- a/content/pages/about/xsf/jabber-trademark/pending-applications.md
+++ b/content/pages/about/xsf/jabber-trademark/pending-applications.md
@@ -35,4 +35,8 @@ Sidebar_menu_elem_url_11: about/xsf/jabber-trademark/approved-applications
 Content_layout: multiple-columns
 ---
 
-Pending applications are now tracked at [http://tracker.xmpp.org/browse/TRADEMARKS](http://tracker.xmpp.org/browse/TRADEMARKS)
+This page lists pending appliations for a license to make use of the [Jabber trademark](/about/xsf/jabber-trademark).
+
+| Mark                                  | Licensee                    | Type                        |
+|---------------------------------------|-----------------------------|-----------------------------|
+|                                       |                             |                             |


horazont:
Is there any reason not to use the github issue tracker instead of having to modify the list on the website all the time?

guusdk:
Not really. The full application holds some privacy-sensitive information, which wouldn't go in the website or github tracker. I'm unsure if that went into the old tracker (or if it was private at all).

horazont:
In my case (<https://github.com/xsf/xmpp.org/pull/358>) I simply emailed the info to trademark@ and created a PR here.

One could craft a similar process where pending applications are instantiated on github either as issue or PR, either by the applicant or by the one handling the application first.

But that’s just a suggestion since it might be easier to manage for Board and the other people involved. I’m not particularly keen on centering more of our process around GitHub.

stpeter:
The process that @horazont suggests would be fine with me. Note that we rarely receive trademark sublicensing requests, so the burden of updating the website seems minor to me.

guusdk:
Do I understand things correctly: we can ditch the 'pending applications' page, and can work with github PRs to the 'accepted applications' page to track whatever application is pending. Works for me.

Ppjet6:
What's the status of this? Does this need board approval? Should we ask @guusdk (or some other volunteer) to remove the page?

SeveFP:
> Do I understand things correctly: we can ditch the 'pending applications' page, and can work with github PRs to the 'accepted applications' page to track whatever application is pending. Works for me.

Hi @guusdk, is this PR still relevant?

Echolon:
Hi @guusdk,

we have been upgrading from Pelican to the Hugo website generator. Could you review this in case it is still relevant?

That would be really great!

guusdk:
The website still does not have a tracker for trademark applications, which in the past, it had. Wether that's still desirable is up for debate.

stpeter:
Trademark applications are so rare these days that I don't think we need a tracker.

Echolon:
@guusdk @stpeter I think we can merge the table, leave it and if we should ever need it its there. But I think it doesn't harm to show it to also keep this topic clear and official.

Echolon:
@guusdk @stpeter

The underlying (HTTP !) website doesnt work really. I would recommend to merge and update with Trademark requests. Everything XSF related should be on our website I assume.

Let me know what to do.

stpeter:
Handling trademark requests via GitHub pull requests is a good idea.

Echolon:
> Handling trademark requests via GitHub pull requests is a good idea.

@stpeter so you meam this ironic and are against it?

stpeter:
I didn't include an ironic emoji, so I meant it seriously. We've tracked trademark requests in many ways over the years, but GitHub would be a good, lightweight, transparent method. Please do note that we very rarely receive trademark requests these days - it's been many years since the last one.

Echolon:
> I didn't include an ironic emoji, so I meant it seriously. We've tracked trademark requests in many ways over the years, but GitHub would be a good, lightweight, transparent method. Please do note that we very rarely receive trademark requests these days - it's been many years since the last one.

Ah okay, yeah sometimes I just dont expect this ^^

I agree with you. Even though it has few requests we are responsible, so let's stick to it.

@wurstsalat3000 please merge is that's alright.
@stpeter feel free add former entries if that makes sense
@stpeter btw please try reply to my messages I sent to your ID :-)

Echolon:
> I didn't include an ironic emoji, so I meant it seriously. We've tracked trademark requests in many ways over the years, but GitHub would be a good, lightweight, transparent method. Please do note that we very rarely receive trademark requests these days - it's been many years since the last one.

Ah okay, yeah sometimes I just dont expect this ^^

I agree with you. Even though it has few requests we are responsible, so let's stick to it.

@wurstsalat3000 please merge is that's alright.
@stpeter feel free add former entries if that makes sense
@stpeter btw please try reply to my messages I sent to your ID :-)

PR #498, opened 2019-01-07
Add a banner for the upcoming summit

horazont:
This is with super fugly inline CSS because I cannot be bothered to install npm.

Screenshot


horazont:
screen shot 2019-01-07 at 21 01 45

PR #499, opened 2019-01-07
Add information about sprints

Ppjet6:


guusdk:
To me, it makes sense to list the 'upcoming' before the 'past' sprints, but I'm happy to merge as-is.

PR #501, opened 2019-01-08
Events: Make conferences a list

Ppjet6:
There was a missing newline between Summit and FOSDEM. I thought a list might be better anyway.


guusdk:
Thanks!

PR #504, opened 2019-01-10
Side menus

guusdk:
Most pages have a menu in a side bar. Once upon a time, this side bar menu was inherited, but that mechanism now fails. To work around this problem, the sidebar definition of a set of pages is copied across many other pages. Over time, modifications sneak in that do not get copied, which makes for a messy structure.

The commits in this PR reflect my journey through discovering how things work. I ended up with replacing the very flexible, but equally complex inheritance-from-an-ancestor implementation to an implementation where the sidebar is inherited from a page that is directly referenced.

### DIFF
--- a/content/pages/about/faq.md
+++ b/content/pages/about/faq.md
@@ -8,22 +8,7 @@ Top_Menu_order: -1
 Dropdown_menu_show: false
 Footer_show: false
 Sidebar_menu_show: true
-Sidebar_menu_title: About XMPP
-Sidebar_menu_size: 7
-Sidebar_menu_elem_name_1: History
-Sidebar_menu_elem_url_1: about/history
-Sidebar_menu_elem_name_2: Overview
-Sidebar_menu_elem_url_2: about/technology-overview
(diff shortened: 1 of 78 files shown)


guusdk:
Note that I've applied the new implementation to the first menu ("About XMPP") only! I'd like some feedback on the approach, before I apply it throughout the site.

intosi:
I think this is a great step in the right direction, thanks for doing this!

guusdk:
@horazont Right you are: https://github.com/xsf/xmpp.org/issues/506

guusdk:
I have finished applying the same fix throughout the website.

horazont:
@guusdk Did you verify that it works with the pelican version for the website? If so, hit the merge and observe whether something breaks afterwards (15 min build time). Revert if it does.

I can’t verify it locally, so I won’t hit merge.

guusdk:
I did. It's pretty easy to run locally with the docker makefile. That pulls in the correct version of Pelican.

PR #507, opened 2019-01-19
Modify banner to fix display on mobile devices

horazont:
Before on mobile:

The banner obstructs the navigation bar

After on mobile:
The banner is below the navigation bar instead of obstructing it.

After on desktop:
The banner is below the navigation bar instead of above the tagline bar


guusdk:
Thanks!

PR #508, opened 2019-01-19
Deliver fonts and JavaScript from the XSF servers instead of Google

karoshi42:
To enhance the privacy of the site visitors, don't tell google they were there. Especially with 'Privacy-focused' being one of the first things which are said about XMPP (and for me definitely one main reason to use it), I think this might be an additional step to support that claim.


.gitignore
```diff
@@ -16,3 +16,5 @@ output/
 bower_components
 node_modules
 content/extensions
+.venv
+package-lock.json
```
guusdk: I don't think package-lock.json is supposed to go in .gitignore.

jcbrand: No, it should go into version control.

Ppjet6:
It seems to me that the banner picture has been left out:

https://raw.githubusercontent.com/adambrault/xmpp.org-static/gh-pages/img/london.jpg

I'm not sure what is blocking this PR otherwise. I'd be happy to accept it as is as it's already an improvement over the status quo.

Echolon:
@karoshi42 have you tried on this again?

I think this is good in terms of privacy.

mwild1:
With the switch to Hugo (#953), all resources are now served from xmpp.org :+1:

PR #517, opened 2019-02-03
Replace list.jabber.at with compliance.conversations.im

404city:
Replacing outdated server catalog with a new. New list XMPP server freedom and independent. Anyone can add their server in a couple of seconds. The new catalog is more good, as it better describes the capabilities of the servers, and not only lists their names.
This directory is more useful for new XMPP users than meaningless enumeration of server names.

### DIFF
--- a/content/pages/getting-started/_index.md
+++ b/content/pages/getting-started/_index.md
@@ -24,7 +24,7 @@ There are [many, many XMPP clients](/software/clients) for you to choose from. T
 
 # 2. Create an account
 
-As with e-mail, you'll need an account with a service provider. Again, there are [plenty of choices](https://list.jabber.at/), from which these are only a couple of the more popular ones:
+As with e-mail, you'll need an account with a service provider. Again, there are [plenty of choices](https://compliance.conversations.im/), from which these are only a couple of the more popular ones:
 
 * [xmpp.jp](https://www.xmpp.jp/signup)
 * [jabber.ru](https://jabber.ru/user/register)


horazont:
Please see my comment on a similar request from the past: https://github.com/xsf/xmpp.org/pull/454#issuecomment-406157898

horazont:
Given that we go out of our way to be neutral regarding which clients we recommend, I am not sure we should endorse a listing where servers are ranked by ~~how determined the admin was to tick all the green boxes~~ implemented XEPs.

Also, a due note: A key difference between the "old" and "new" listing as proposed in this PR is that 404.city is not listed on the old, and listed first on the new. In that context, I’d like to remind [of the discussion we had last year](http://logs.xmpp.org/xsf/2018-07-05/#14:03:32).

404city:
> Given that we go out of our way to be neutral regarding which clients we recommend, I am not sure we should endorse a listing where servers are ranked by how determined the admin was to tick all the green boxes implemented XEPs.


You are trying to maintain independence, but the list of servers list.jabber.at depends on the administrator jabber.at. Administrator jabber.at adds or removes servers according to the principle that they don’t like it. These are non-binding criteria.

404city:
> Given that we go out of our way to be neutral regarding
I ask only about an independent list, I do not do advertising.We need help, not preconceptions.

SamWhited:
Jabber.at seems as good a source as any. I'd rather have a list that's had at least minimal vetting too.

I'd also argue that we shouldn't use the compliance checker list just because it has lots of confusing XEPs on it that most people won't understand or care about, while the jabber.at one just lists servers and where they are.

404city:
> Jabber.at seems as good a source

Jabber.at will remove the servers from the list, as desired. If the server is not liked by the Jabber.at administrator, it will remove the server. I want an independent list

SamWhited:
> Jabber.at will remove the servers from the list, as desired. If the server is not liked by the Jabber.at administrator, it will remove the server. I want an independent list

That seems fine. A vetted list isn't a bad thing.

404city:
> That seems fine. A vetted list isn't a bad thing.

jabber.at deleted the 404.city server due to a personal conflict. I find it inadmissible to advertise a list that has preconceived criteria for adding or removing

SamWhited:
> jabber.at deleted the 404.city server due to a personal conflict. I find it inadmissible to advertise a list that has preconceived criteria for adding or removing

What sort of "personal conflict"?

guusdk:
The 'getting started' page, where we link to the list, is intended for users who are completely unfamiliar with XMPP. As the jabber.at list links to 'how to register on this domain' type of pages, it provides 
very helpful information for exactly those users. Although the compliance checker is an excellent tool, I suspect it will confuse this intended audience more than it will help. I see no compelling reason to remove the jabber.at list. I'd welcome adding additional lists, but do not think that the compliance tester list is suitable.

404city:
> 404.city was removed from my list because the admin falsely reports users as "spam" (when in reality they just like Matrix over XMPP) and explicitly stated (several times) that he will block any user on his server that prefers Matrix over XMPP.
> 
> The list is a directory of servers that you can safely use. If an admin blocks you just because he doesn't like Matrix, that's not safe to use.

404.city was removed from my list because the admin falsely reports users as "spam" (when in reality they just like Matrix over XMPP) and explicitly stated (several times) that he will block any user on his server that prefers Matrix over XMPP.

The list is a directory of servers that you can safely use. If an admin blocks you just because he doesn't like Matrix, that's not safe to use.

PR #520, opened 2019-02-04
Add a header-banner CSS class for the next summit or event

linkmauve:
This is a follow-up to #519, to remember the banner’s CSS values for the next time we decide to set one.

Use `<a class="header-banner" href="/community/events/summit24">Blah blah blah</a>` for the 24th summit for instance.

### DIFF
--- a/xmpp.org-theme/sass/components/header.sass
+++ b/xmpp.org-theme/sass/components/header.sass
@@ -25,3 +25,11 @@
 
 .header-internal h1, .header-internal h1 a
   color: $dkblue
+
+.header-banner
+  color: white
+  font-size: 120%
+  background: #008CBA
+  padding: 20px
+  display: block
+  clear: both


horazont:
I *think* you have to compile the sass, too. This is why I didn’t do it via proper CSS in the first place.

linkmauve:
Isn’t that the role of the deployment script?

horazont:
I did not find any code which actually does that.

SeveFP:
@linkmauve any updates on this one?

mwild1:
@wurstsalat3000 does the hugo branch contain something for this, or do we need to port it?

wurstsalat3000:
@mwild1 yes it does! I added some CSS and furthermore, I added `data/announcement.json`. That file allows to enable/disable and set an announcement text including a URL.

PR #534, opened 2019-02-28
Renewing Simple IoT Client entry

PeterWaher:

PR #535, opened 2019-02-28
Renewing Simple IoT Client entry

PeterWaher:


guusdk:
Am I missing something, or is this a duplciate of #534 ?

guusdk:
Ah, there's a different year in the renewal date.

PR #549, opened 2019-04-03
Added rebranded Tigase Messenger for iOS as Siskin IM

hantu85:
Added rebranded Tigase Messenger for iOS as Siskin IM and used full client names for Siskin IM and BeagleIM.

Only iOS client was rebranded. Tigase Messenger for Android kept its name for now.


guusdk:
Thanks!

PR #563, opened 2019-05-07
Update clients.json

damkh:
Supported platforms of Kaidan

### DIFF
--- a/data/clients.json
+++ b/data/clients.json
@@ -306,10 +306,7 @@
         "name": "Kaidan",
         "platforms": [
             "Android",
-            "iOS",
             "Linux",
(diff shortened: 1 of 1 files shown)


guusdk:
This seems to correspond with the download options as provided on https://www.kaidan.im/download/

Thanks!

JBBgameich:
Unfortunately this is not entirely correct, since Kaidan can run on iOS and MacOS even though we don't have any official builds yet. It's nevertheless possible to compile Kaidan for both Platforms.
I believe especially the removal of the "Other" platform is unjustified, as this was probably meant to show the support for Ubuntu Touch.

PR #568, opened 2019-05-18
Update dates

wrooot:
I was under impression that a client is shown in the list if its renew date is fairly recent. Or is there another mechanism? Say, Pidgin is not on the list, its renew date is 2018-03-31. Maybe this date should be not older than a year for a client to show up?

Also, it seems that renew date should be a date when someone submits a PR with an update for that client. But i have just added dates of actual last release for some more or less known clients (ChatSecure is not showing on the list, but it had a release a month ago, so i think it should be there).

### DIFF
--- a/data/clients.json
+++ b/data/clients.json
@@ -96,7 +96,7 @@
         "url": "https://candy-chat.github.io/candy/"
     },
     {
-        "last_renewed": "2017-04-19T14:37:00",
+        "last_renewed": "2019-04-22T14:37:00",
         "name": "ChatSecure",
         "platforms": [
             "iOS"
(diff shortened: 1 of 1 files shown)


guusdk:
Hi! Thanks for your efforts, but they might be somewhat overzealous.

The date that we ask you to fill out is explicitly _not_ the "last release" date of the software. It's the date that someone from the project or community has bothered to update the listing of that project on our site (basically, when they last submitted a PR to this repo).

The idea is that, by requiring people to update their listings once a year, the projects that are unmaintained/unused will automatically disappear from our list (we used to list a _lot_ of very old, incompatible, broken software - this process intends to reduce that, and has been pretty successful at that, in my opinion).

You can find more background in this article that we posted at the time we implemented this process: https://xmpp.org/2017/03/new-xmpp-software-listing-rules/

wrooot:
I see.

PR #570, opened 2019-05-22
Add Karlsruhe and The Hague sprints and show banner for them

wurstsalat3000:
This adds both the Karlsruhe sprint and the The Hague sprint to the community events page.

Additionally, a banner is shown on the front page (with a link to community events).


guusdk:
Thanks!

PR #574, opened 2019-06-11
Updated platforms.json and added UWPX

COM8:
I added [UWPX](https://uwpx.org/) to the `clients.json`.

I also added `Windows 10` to the `platforms.json`.
Why:
* UWPX is a Windows 10 client app that runs only on Windows **10** devices not on for example Windows 7/8/...
* Windows 10 as a platform includes: Xbox, HoloLens, Windows Phone, PC, ...

### DIFF
--- a/data/clients.json
+++ b/data/clients.json
@@ -565,6 +565,14 @@
         "url": "https://trillian.im"
     },
     {
+        "last_renewed": "2019-06-11T01:10:47",
+        "name": "UWPX",
+        "platforms": [
+            "Windows 10"
+        ],
+        "url": "https://uwpx.org"
+    },
+    {
         "last_renewed": null,
         "name": "V&V Messenger",
         "platforms": [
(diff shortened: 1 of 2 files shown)


guusdk:
Hmm... I'm not to fond of adding new platforms (this has been discussed before - it muddies the water, without adding much value to the average user - people that are using very specific platforms typically already know where to get their resources). That said, UWP might be large enough to warrant inclusions. (why not add "UWP" instead of "Windows 10"?)

Thoughts?

COM8:
I don't think `UWP` would be a good platform name since it will probably change soon and if I ask people that are not that tech savvy what `UWP` is, they say: "Never heard of it. What is UWP?"

This is the reason why I think `Windows 10` would be a good platform name (at least until the support for Windows 7-8.1 runs out).

Echolon:
Stick to Windows, is fine!

COM8:
OK switched back from `Windows 10` to `Windows`.

COM8:
Looks like Travis can't find the python 3.3 package.
Can somebody restart the CI?

PR #575, opened 2019-06-11
Remove Karlsruhe and The Hague sprint banner

wurstsalat3000:
This moves both the Karlsruhe sprint and the The Hague sprint to the
past events section (events page). Additionally, this removes the banner
on the front page.


guusdk:
Thanks!

Issue #585, opened 2019-07-03
Software comparsion tables for supported XEPs


Mikaela:
IRCv3 has nice comparsion tables that name the clients/servers/libraries and what IRCv3 extensions they support. I think the idea would port over well to XMPP as XEPs work similarly to IRCv3 extensions, but as there are so many of them, I guess just the compliance suites that users are going to see would be enough to be listed.

* https://ircv3.net/software/clients
* https://ircv3.net/software/servers
* https://ircv3.net/software/libraries

IRCv3.net is using Jekyll and the source is at https://github.com/ircv3/ircv3.github.io .

ge0rg:
I suggest using the Compliance Suites for that, or maybe at least having an aggregate "Compliance Level" column for implementations, because the individual XEPs are very many and very noisy for normal users.

wurstsalat3000:
https://xmpp.org/software/ can now be used to filter by Compliance level and by individual XEPs. I'll close this issue in favor of #1185

PR #594, opened 2019-07-27
Add a DOAP URL for each implementation we know of

linkmauve:
I originally introduced my DOAP proposal at standards@ about two years ago: https://mail.jabber.org/pipermail/standards/2017-August/033123.html

This semantic web format describes a project, I extended it with information about XEP support, most relevant to XMPP software. All providers I know of are using its XML serialisation, but other serialisations exist such as Turtle or JSON-LD, they only need support in the triple parsing library used.

I’ve written an implementation in Python, attached in this pull request, as well as a JavaScript one which can be tested here: https://linkmauve.fr/extensions/xep-0048.xml

The Python one imports all of the DOAP information to Pelican, but doesn’t do anything with it yet. It would be nice to store it locally on the server to avoid remote calls at other times than server creation. It would also be useful to generate a per-XEP JSON file or similar, containing only the information they need to render.

I would welcome a better solution than modifying the DOM from JavaScript for XEP integration, but my attempts with using XSLT with XInclude didn’t reach anything usable.

This could be used to implement #585 in the future.

### DIFF
--- a/data/README.rst
+++ b/data/README.rst
@@ -38,7 +38,7 @@ The tool will ask for confirmation::
    }
   is this okay? [y/n]
 
-After confirmation, it writes the changes to the ``clients.json``. This works just the same for ``servers.json`` and ``libraries.json``. You can then add and commit the changes to git as usual. **Validate** that your entry is correct using the ``./lint-tool.py`` on the respective JSON file and then make a Pull Request on GitHub.
+After confirmation, it writes the changes to the ``clients.json``. This works just the same for ``servers.json`` and ``libraries.json``. You can then add and commit the changes to git as usual. **Validate** that your entry is correct using the ``./lint-list.py`` on the respective JSON file and then make a Pull Request on GitHub.
 
 
 Updating information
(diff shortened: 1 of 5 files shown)


Ppjet6:
What happens with this? Should it be submitted to board?
Is there anything that needs to be changed implementation-wise?

jcbrand:
> Travis seems unhappy

Looks like it fails trying to build with Python 2.7 (while passing for newer versions). Do we still care about Python 2?

I restarted the job, maybe it's an intermittent issue.

jcbrand:
I think @linkmauve should rebase this branch onto master, then the error might likely go away.

linkmauve:
It seems the issue is with python2.7.

I think it would make sense to drop python2 support altogether, as it has only three months left to live.

linkmauve:
This PR is currently blocked on #621.

wurstsalat3000:
I'd like to add that Gajim's DOAP file is available here: https://dev.gajim.org/gajim/gajim/blob/master/data/gajim.doap

Ppjet6:
@wurstsalat3000, I suggest we add this once the PR gets merged, (that is, once it's unblocked..).

linkmauve:
I guess I could also remove the Python changes from this PR, and open a new one with them, that way you can merge it.

linkmauve:
I just removed the Python parts altogether, leaving only the addition of a DOAP field in the JSON files.

This should be ready for merging.

Ppjet6:
@wurstsalat3000, this has been merged, you should be able to PR for the gajim entry now :)

wurstsalat3000:
It's already included I just saw :)

PR #604, opened 2019-08-30
update memberlist

agnauck:
update memberlist after Q3-2019 elections


ralphm:
The changes look good. I think the build fails because it relies on the ancient Python 3.3, which apparently can no longer be downloaded over at TravisCI. Maybe this requires a closer look.

Please merge.

PR #612, opened 2019-09-20
Please add txxmpp to clients.json

dd9jn:
Due to missing perl libraries in newer Debian versions and other problems with sendxmpp I wrote new new tool named txxmpp.c based on the widely available strophe library.  We use it to push commit announcements to the gnupg-devel MUC.  It is a plain and simple tool with all instructions in the help output.  I am not sure whether it fits into the client list but I have not seen a better place, yet.

### DIFF
--- a/data/clients.json
+++ b/data/clients.json
@@ -647,5 +647,13 @@
             "iOS"
         ],
         "url": "https://zom.im/"
+    },
+    {
+        "last_renewed": null
+        "name": "txmmp",
+        "platforms": [
+            "Unix" 
+        ],
+        "url": "https://git.gnupg.org/cgi-bin/gitweb.cgi?p=gnupg-doc.git;a=blob;f=tools/txxmpp.c"
     }
 ]
--- a/data/clients.json
+++ b/data/clients.json
@@ -650,7 +650,7 @@
     },
     {
         "last_renewed": null,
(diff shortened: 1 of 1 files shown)


guusdk:
I'm thinking that this is more of a tool/library than a client, in context of usability to regular end-users. Also, is there any kind of documentation on this, other than the source code at https://git.gnupg.org/cgi-bin/gitweb.cgi?p=gnupg-doc.git;a=blob;f=tools/txxmpp.c ?

Echolon:
Hello @dd9jn,

May you have a response to the previous question?

There is also a libraries section on the XSF website: https://xmpp.org/software/libraries.html

dd9jn:
> May you have a response to the previous question?

It is an xmpp client in the form of a standard Unix tool.  Nope, there is no more documentation than the command line help output whcih also gives an example.

Echolon:
I added two fixes, however I don't know why it does not build now ...

Thank you for the contribution btw!

SeveFP:
@dd9jn I can't merge it until all checks are green. Can you take a look, please? I'll merge as soon as this gets fixed.

Echolon:
I will close this with reference to this: #612

I hope that work. Sorry, for taking ages!

PR #613, opened 2019-09-25
Update profanity website

mdosch:
Profanitys website moved from profanity.im to https://profanity-im.github.io/.
Right now there is a forwarding set up, but they are considering to ditch
that domain.


guusdk:
Thanks!

PR #617, opened 2019-09-30
Fix markdown link syntax

mdosch:


mdosch:
I did something wrong it seems...

PR #621, opened 2019-10-04
Update Python version to 3.8

linkmauve:
The website was still built under Python 2.7, which is due for EOL in
less than three months.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,17 +5,17 @@ notifications:
 
 jobs:
   include:
-  - python: 2.7
+  - python: 3.8
     name: "Website Build"
-  - python: 3.7
+  - python: 3.8
     name: "Data Linting"
 
 install:
-- if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then pip install pelican==3.3 ghp-import markdown; fi
+- if [[ $TRAVIS_PYTHON_VERSION == 3.8 ]]; then pip install pelican==3.3 ghp-import markdown; fi
 
 script:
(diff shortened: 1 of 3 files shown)


horazont:
This is broken:

> Done: Processed 0 articles and 1 pages in 0.51 seconds.

horazont:
Hm, okay, the output is a red hering. In fact, it generates a lot of output. The output is broken however (the classic nav bar bug). It is unclear whether this would be fixed by a build on the actual build server.

I have never figured out how to produce a valid build locally.

guusdk:
@horazont What's a pragmatic approach? Give it a try and roll back if it doesn't work?

Zash:
Adding a `chmod a+r /var/www/html -R` in the Dockerfile seems to fix the weird permission issues.

linkmauve:
Ok, testing that suggestion then.

Echolon:
Hi @linkmauve,

may you have time again to review this?
That would be really great!

wurstsalat3000:
> Hm, okay, the output is a red hering. In fact, it generates a lot of output. The output is broken however (the classic nav bar bug). It is unclear whether this would be fixed by a build on the actual build server.
> 
> I have never figured out how to produce a valid build locally.

@horazont could you try this in conjunction with #743 ? This should enable you to test locally using a recent Pelican version. I managed to build the site locally using Pelican 4.2 and Python 3.8.

mwild1:
This is obsolete now that we have switched to Hugo.

PR #622, opened 2019-10-05
getting-started: linkify "visit one of the many chatrooms"

Mikaela:
I think the link to https://search.jabber.network/ was missing as the page doesn't otherwise have any hint on where to find the many chatrooms.


guusdk:
Thanks!

PR #626, opened 2019-10-24
Blogpost: XSF Board and Council 2019-2020 Elections

guusdk:


guusdk:
When running locally, opening the 'read entire blog post' link gives me a 404. I'm not sure if that's me, or something wrong with the post. Can someone verify this please?

content/posts/blog/2019-elections-call-for-candidates.md
```diff
@@ -0,0 +1,20 @@
...
+title: XSF Board and Council 2019-2020 Elections
+date: 2019-10-24 10:54
+author: guus
+blog_id: blog
+category: XSF Organisational
+---
+
+The XMPP Standards Foundation will elect the 2019-2020 XSF Board and Council soon. Each year the members of the XSF choose new people to provide technical leadership on the [XMPP Council](https://xmpp.org/about/xmpp-standards-foundation.html#council) and business leadership on the [XSF Board of Directors](https://xmpp.org/about/xmpp-standards-foundation.html#board).
+
+For the upcoming election, the XSF is looking for candidates for both roles! Candidates for the XMPP Council are searched for within the ranks of the XSF Members, while the role of Director of the XSF Board is open to anyone. Both roles typically require only a few hours a week over the one-year term, and are a great way to get involved in the XMPP community.
```
Ppjet6: > Both roles typically require only a few hours a week over the one-year term, and are a great way to get involved in the XMPP community.

I read this as something that can be taken lightly as not really important. Both roles can actually take a lot more time than "a few hours". Maybe this can be removed or reworded. It's Board and Council member's job to make their role interesting, to make it matter.

guusdk:
> When running locally, opening the 'read entire blog post' link gives me a 404. I'm not sure if that's me, or something wrong with the post. Can someone verify this please?

I'm now noticing that this is also happening (locally) with older posts. Those posts are fine on the live website, so it's probably something with the local setup.

content/posts/blog/2019-elections-call-for-candidates.md
guusdk: From personal experience, a few hours a week seems fair, for Board membership anyways. I've not been on Council, so I can't share experiences there.

I'd love to avoid creating the impression that is to be taken lightly, but on the other hand, I do think it is good to give some kind of indication on what amount of work people would be volunteering for.

I'm open to improving this text, but I have a hard time coming up with a better alternative. Suggestions?

PR #628, opened 2019-10-26
update-entry: Fix missing newline at the end of generated files

lnjX:


guusdk:
Thanks, but why?

lnjX:
@guusdk Most editors add a newline at the end of a file. So in case the file is edited manually a newline is added and would then be removed again the next time the script is used. This just messes up the git history. I noticed it happening here already (and also had this "problem" in other projects).

guusdk:
I have zero confidence that this will actually improve anything - but I don't think that it will make anything worse either. Given that no-one else voiced a concern, let's give it a try.

horazont:
@lnjX Thanks, those diffs have been driving me crazy.

PR #630, opened 2019-11-06
Bylaws: Allow electronic communications to request special meetings

Ppjet6:
Specify that electronic communications are allowed to request special meetings in §3.3.


Ppjet6:
Indeed. I am curious though why it's specified in so many other places though, but meh. Thanks!

PR #633, opened 2019-11-12
Uses/IM: correct nimbuzz link, old one is 404

ge0rg:


guusdk:
Hmm, I'd be very cautious to refer to this page. As far as I know, Nimbuzz has ceased to exist completely. I'm not sure who is serving what content at the domain that you're adding here.

ge0rg:
Indeed, all I did was a cursory look and it seemed at least to match the spirit. Should we remove Nimbuzz altogether, then?

guusdk:
Yes, I think so. There's little value in showing off a use-case that doesn't exist any more.

Ppjet6:
Can this be closed then?

ge0rg:
Alright, I've just killed Nimbuzz completely. Feel free to apply PR if appropriate :)

guusdk:
Works for me! Thanks!

Issue #635, opened 2019-11-14
Update history


Neustradamus:
The history is stopped in 2017: https://xmpp.org/about/history.html.

Nyco:
Fix: #636

Nyco:
Fixed. Thanks!

PR #636, opened 2019-11-14
2019 and 2020 summits

Nyco:
We forgot to update for 2019, planned for 2020.
Fixes issue: #635


SeveFP:
Thank you @Nyco 
Those years look a bit empty, I wonder if we can write a bit of progress on each of those years.

Issue #566, opened 2019-05-14
Alphabetic ordering in software listing


guusdk:
See #565: The existing code orders "Pàdé" behind "Pidgin". That aught to be reversed.

Nyco:
Web page: https://xmpp.org/software/clients.html
Git page: https://github.com/xsf/xmpp.org/blob/master/content/pages/software/clients.md

How do we update the content? It is not in the markdown file...

guusdk:
> How do we update the content? It is not in the markdown file...

See https://github.com/xsf/xmpp.org/blob/master/data/README.rst

Nyco:
Thx @guusdk !

PR #637 should fix this issue, please review.

Echolon:
As the fix has been merged I will close this!
Thanks!

PR #637, opened 2019-11-14
Pade before Pidgin

Nyco:
Alphabetical order
Should fix issue #566


horazont:
The tooling is currently not unicode-aware. It orders strictly by codepoint number.

To fix this, we’d have to include something like libidn or so.

Zash:
What if it's decomposed into `Pa\u00300de\u0301` ?

Nyco:
So I'd close this one... Agree?

PR #641, opened 2019-11-20
CommTeam & SCAM chats?

Nyco:
The CommeTeam was not listed.


guusdk:
Does it add value to the general public to know about this MUC? I'm not saying it's a secret or anything, but I'm thinking that no-one outside of the XSF has interest in this. If that's true, then adding it to the list of MUCs intended for the general public might add more confusion than be of benefit. The same probably goes for most other WorkTeam-related chats. If we do want to list them, then I suggest we create a new section for "mostly XSF-internal chats, but feel free to pitch in if you're interested" type of things.

Nyco:
Good point...

Nyco:
Discussions happened on the CommTeam MUC: we chose not to mention on internal work team on this page.

PR #649, opened 2019-11-21
artalk-xmpp

ammarheidari:
C# library for XMPP

### DIFF
--- a/data/libraries.json
+++ b/data/libraries.json
@@ -25,7 +25,7 @@
     }
 ,
     {
-        "last_renewed": null,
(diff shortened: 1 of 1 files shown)


agnauck:
why did you open another pull request for this?

guusdk:
Thanks for your contribution. You seem to be struggling a little with the format in which this data is to be submitted. For future reference, it's documented here (also lists a tool that will automatically generate the correct format): https://github.com/xsf/xmpp.org/blob/master/data/README.rst

Your contribution didn't include a timestamp, which means that it won't be listed on the website. I've added timestamp for you, and dated it to yesterday.

PR #655, opened 2019-11-23
add RFC 7590 to extensions

sonnyp:


guusdk:
At some point, we might want to present this in a more visible manner (a listing of sorts?) to put a bit more emphasis on these RFC - but for now, the proposed addition in itself is a nice improvement. Thanks!

sonnyp:
@guusdk right, other XMPP RFCs are missing. Also the page title is "Specifications" but the url is `/extensions`

PR #661, opened 2019-11-29
libraries: Fix QXmpp link

lnjX:
http://qxmpp.org is currently unavailable (@jlaine ?)


jlaine:
@lnjX thanks for noticing this. I've opened https://github.com/qxmpp-project/qxmpp/issues/261 to handle transfer of the domain

PR #663, opened 2019-12-07
Add DOAP entry for Prosody

Zash:
The tool sorted all the fields so I committed that too because why not?

[A manually rendered version of the DOAP file](https://prosody.im/files/prosody-doap.html)


Ppjet6:
Thanks!

PR #683, opened 2020-02-26
Add ObjXMPP to libraries.json

Midar:


jcbrand:
@Midar Why do you link to the Github mirror and not the actual repo? https://git.nil.im/objxmpp.git

Echolon:
Hi @Midar,

do you have a response to the question above?

Midar:
Sorry I missed it.

The reason is that git.nil.im is going away. If anything, https://fossil.nil.im/objxmpp would be a better place (that's the master that creates the GitHub repository), but I think for most people, GitHub is what they're used to and hence makes more sense.

PR #686, opened 2020-03-03
Add Grindr as project that uses XMPP for "instant messaging"

jcbrand:
Also, remove chatme.im and jappix, both of which are no longer active.


deltachonglee81 (commented):
Good work

PR #695, opened 2020-04-10
Update jQuery to the maintained 3.5.0 version

### LINKED ISSUE #694: Remove Google's APIs and other google assets from xmpp.org (privacy reasons)
It's deeply saddening that xmpp.org is relying on google's assets, since we all know very well how privacy-invasive Google is and how little they care about us having an open internet. They wanna censor and control as much of the internet as they can.

linkmauve:
This also pulls the JavaScript file into our repository, to avoid a dependency on googleapis.com.

I haven’t seen any regression for the top bar, which is AFAIK the only thing we use jQuery for.

Fixes #694.


jcbrand:
@guusdk When I just tested with jQuery 3.5.0 for #696 I saw jQuery related errors. Which is why I decided to stick with jQuery 2.

PR #699, opened 2020-04-12
Remove dead link

qaisjp:


linkmauve:
As an alternative, you could use the [Wayback Machine](https://web.archive.org/) to get the latest indexed version of this page: https://web.archive.org/web/20160531172428/http://www.xmppresearch.org/posts/xeps-used-in-xmpp-research/

PR #701, opened 2020-04-18
Patch 8

mdosch:


mdosch:
Oops, this should only go against the newsletter PR. Sorry for the noise.

PR #707, opened 2020-05-06
Update clients.json

resto1231:
Add mailfence to the list of clients. It has an XMPP for users to chat with their contacts.

### DIFF
--- a/data/clients.json
+++ b/data/clients.json
@@ -362,6 +362,15 @@
     },
     {
         "doap": null,
+        "last_renewed": "2020-06-05T10:58:39",
+        "name": "mailfence",
+        "platforms": [
+            "Other"
+        ],
+        "url": "https://mailfence.com"
+    },
+    {
+        "doap": null,
         "last_renewed": "2019-07-16T20:20:38",
         "name": "mcabber",
         "platforms": [


Ppjet6:
Hi, thanks for your PR!

Having a look at your website I am curious whether this is an XMPP client or if you only propose XMPP services: https://kb.mailfence.com/kb/could-you-please-suggest-any-local-client-for-xmpp-connection-for-mailfence-group-chat/. Is this link only suggesting alternative ways to connect and you already propose one by default?

I also couldn't find anything regarding licenses, can you help here?

I was wondering if it might fit better in https://xmpp.org/uses/

Neustradamus:
@resto1231: Maybe time to update the website, no information about XMPP on homepage...

And I see in the previous link "Pidgin", but it is not a good XMPP client.

Please change with Gajim and Psi for example.

resto1231:
Thank you all for the incredible response. I have notified my team and we are working on it to provide you the required info. Will update you on the coming days

I should have asked before making a pull request. I will be more cautious next time.

resto1231:
Hello Guys,

One final thing before I close this request as the tech team replied and said indeed Mailfence is not fit for the clients' list. Thank you again for the recommendations the tech team has updated the KB article :)
This list here could potentially apply more to Mailfence's scenario. Should I open a request for the below list?
https://xmpp.org/uses/social.html

Echolon:
@resto1231 this thread can be closed?

PR #729, opened 2020-07-02
Review and GSoC additions

wurstsalat3000:
@Echolon


Echolon:
Thank you very much for your review as well as the additions to GSoC!

PR #743, opened 2020-07-16
Pelican: Fix usage of PAGES variable

### LINKED ISSUE #506: Pelican compatibility
The current website depends on Pelican version 3.3. When a newer version is used, the menu is not generated properly.

The website should work with up-to-date versions of Pelican.

wurstsalat3000:
See: https://docs.getpelican.com/en/stable/faq.html#since-i-upgraded-pelican-my-pages-are-no-longer-rendered

Fixes #506 

~Pelican documentation for v3.3 also lists a lowercase 'pages' variable, so it should work with the currently used Pelican version as well, see https://docs.getpelican.com/en/3.3.0/themes.html#common-variables~

AFAIK we are stuck on Pelican 3.3 because the menu went missing as soon as we tried to update to a newer Pelican version. This PR does not work with Pelican 3.3 (currently used), Pelican needs to be updated to a recent version. I tried with 4.2, and that worked well (menu shown, no issues found).

### DIFF
--- a/xmpp.org-theme/templates/footer.html
+++ b/xmpp.org-theme/templates/footer.html
@@ -18,7 +18,7 @@
           Home
         </a>
       </li>
-      {% for page in PAGES|sort(attribute='footer_order') if page.footer_show|default("false") == "left" %}
+      {% for page in pages|sort(attribute='footer_order') if page.footer_show|default("false") == "left" %}
       <li>
         <a href="{{ SITE_URL }}/{{ page.url }}">
           {{ page.menu_name|default(page.title) }}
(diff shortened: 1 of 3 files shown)


jcbrand:
Looks fine, but I haven't tested it. Are you sure it works with the currently used version of Pelican as well?

wurstsalat3000:
I just tested this PR via Vagrant. The menu is missing using Pelican 3.3.

It would be nice if someone could test this PR using a current Pelican version (e.g. 4.2). If this works correctly, we could get rid of the pinned Pelican version.

mwild1:
This is obsolete now that we have switched to Hugo.

PR #744, opened 2020-07-22
End-user friendliness, pass 2

pmaziere:


Echolon:
Thanks for your efforts. I will merge this. However, keep in mind that this is a newsletter. Futhermore we should highlight the important points, even risking the information is not valuable for everyone.

Echolon:
I think we should also always link the first author to review changes @erszcz

erszcz:
I'm fine with these changes. Thanks for taking time to improve the existing text!

PR #745, opened 2020-07-22
Add Monal betas for 4.7

pmaziere:


Echolon:
Really good! Can you just name his full name and write the complete word (nav to navigation)?

PR #748, opened 2020-07-24
Add information about Tigase XMPP Server 8.1.0 release; correct JaXMPP release link.

woj-tek:


Echolon:
Thank you very much!
But with regard to the newsletter, I think we might need to discuss the length 🤔 Still, I also dont want to limit the efforts of the community! 🤷‍♂️ Great work!

woj-tek:
We can just trim it to title + link. I was just following the comment: "_To everyone involved, please do not just drop links, rather try to make a short and decent formulated description of what is the main outcome of your source. This will help us a lot! ". Feel free to cut/trim it as you like :-)

Echolon:
Thanks, is fine. Yes, true that sounds contrary. I did that because previous times, many people just pasted links or only very short describtions or very formalised texts.

Is fine for the moment, we will find the good amount of text by time! Its the first time people contribute such sizes of texts :) But I think that is a good sign!

Jul 27, 2020 3:22:20 AM Wojtek <notifications@github.com>:

PR #749, opened 2020-07-26
Newsletter Aug. 2020: include what-about-design article

Ppjet6:


Echolon:
Thanks!

PR #757, opened 2020-08-02
Monal 4.7, eturnal fixes

licaon-kter:


Echolon:
Very good, thank you!

PR #762, opened 2020-08-05
Trademark - Remove .html from Save_as:

### LINKED ISSUE #203: Various issues with the trademark pages

Echolon:
Is it done with this "simple" change?

@ge0rg

closes #203

### DIFF
--- a/content/pages/about/xsf/jabber-trademark/trademark-license-agreement.md
+++ b/content/pages/about/xsf/jabber-trademark/trademark-license-agreement.md
@@ -1,7 +1,7 @@
 ---
 Title: Jabber Trademark License Agreement
-Url: about/xsf/jabber-trademark/trademark-license-agreement
-Save_as: about/xsf/jabber-trademark/trademark-license-agreement
+Url: about/xsf/jabber-trademark/trademark-license-agreement.html
+Save_as: about/xsf/jabber-trademark/trademark-license-agreement.html
 Parent_id: about
 Top_menu_show: false
 Top_Menu_order: -1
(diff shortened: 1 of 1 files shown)


ge0rg:
I don't know enough about the pelican build to estimate that, and I don't have a local deployment around to test. Sorry, no idea.

Echolon:
@Zash may you have a look at the commit? Is is correct?

Zash (changes requested):
I don't think this is correct. The `.html` suffix should probably be there, as it is on every other page.

Echolon:
@Zash you mean like this? I added .html to both. maybe that was the error?

mwild1:
This is obsoleted by the switch to hugo. The linked issue remains open (I added a comment there).

PR #769, opened 2020-08-25
[Newsletter] Update XEPs

wurstsalat3000:
@Echolon


Echolon:
Thanks for the update!

PR #779, opened 2020-08-31
Renew profanity

Ppjet6:
@jubalh, can I get an ACK please :)

### DIFF
--- a/data/clients.json
+++ b/data/clients.json
@@ -445,7 +445,7 @@
     },
     {
         "doap": null,
-        "last_renewed": "2019-07-16T20:28:37",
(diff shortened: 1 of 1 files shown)


jubalh:
@Ppjet6 actually I have no idea what this is about :-)
Can you explain to me even though it got already merged?

Ppjet6:
It makes you appear in the following list: https://xmpp.org/software/clients.html

Ppjet6:
See also https://xmpp.org/2017/03/new-xmpp-software-listing-rules/

jubalh:
I see. Thanks for doing this!

PR #780, opened 2020-08-31
[Newsletter: August] Fixed the XMPP job board link

COM8:


Echolon:
Well spotted! :eagle:

PR #790, opened 2020-09-05
Fix 'Jobs' being highlighted when viewing 'Blog'

Zash:
<https://xmpp.org/blog.html>

Before:

xmpp org-before
After:

xmpp org-after

Looks like a simple copypaste mistake.


SeveFP:
Thank you very much @Zash ! :trophy:

PR #798, opened 2020-09-09
[august newsletter fix] Attempt to fix HTML rendering of unordered lists

pmaziere:
adding a empty line before the list as well as 2 spaces at the end of each items… not sure about what I'm doing here


Echolon:
Lets take a look in Schrödingers cat! 😄

PR #802, opened 2020-09-14
Translation was added

RincewindsHat:
I translated the article. It is available in english now, too.


Echolon:
Perhaps you rewrite that there is a German translation now as an update in the next release as well?

RincewindsHat:
> Perhaps you rewrite that there is a German translation now as an update in the next release as well?

I am sorry, but I don't quite understand the sentence. Could you elaborate a bit?

Echolon:
I was just suggesting to announce your German article in the upcoming September release

PR #803, opened 2020-09-16
Add libstrophe 0.10.0 release

mdosch:
@pasis  @Echolon


Echolon:
Thanks! May you write something like: LibStrophe, the library that does X Y Z, was released...?

We intend to show readers a bit on the background of most software. Can also just described if it does something in a specific language.

Echolon:
Really good!

PR #816, opened 2020-10-07
[Newsletter October] Update XEP section

wurstsalat3000:
@Echolon


Echolon:
Thank you very much! Those are really important contributions!

PR #818, opened 2020-10-11
Update header image source

SeveFP:


Zash:
The `.sass` file is not actually used, this did not fix the issue. The real path to this background image is in a minifed `.css` file. My attempt to rebuild the `.css` file from the `.sass` resulted in several screens full of npm errors and C++ compilation errors and then I gave up.

PR #823, opened 2020-10-23
Add editor role to Martin Dosch

horazont:
As per Board minutes: https://mail.jabber.org/pipermail/members/2020-October/009194.html


SeveFP:
Thank you!

PR #836, opened 2020-11-16
Add LE certificate announcement

Echolon:
Hello,

thanks to help of @ge0rg I could prepare this PR.

I would be happy for another opinion

@ralphm  ?


ralphm (approved):
I am a bit hesitant about suggesting to bundle certificates in general, but I think the way this is worded, is acceptable. Thanks!

Echolon:
> I am a bit hesitant about suggesting to bundle certificates in general, but I think the way this is worded, is acceptable. Thanks!

Thanks reviewing! How would you change the wording?

PR #842, opened 2020-12-03
history: Add link to this old i-d I found

Zash:
I went digging around with the Wayback Machine and found references to
this internet-draft, and lo, it's still in the IETF document system!

This seems to predate both the XMPP name and the Core-IM split.


Echolon (approved):
Neat! Detective potential detected!

SeveFP:
Awesome!
Thank you @Zash :trophy:

PR #843, opened 2020-12-04
Libraries: add DOAP file for slixmpp

mathieui:
(and renew as a side effect)


Echolon (approved):
Thanks!

Echolon (commented):
Ehm, -> the link does not work?

mathieui:
@Echolon sadly the whole gitlab instance is down at the moment, but otherwise this is the correct URI!

Echolon (approved):
Alright, thanks!

PR #849, opened 2020-12-09
Add txxmpp to clients.json

Echolon:
Quote from Werner Koch @dd9jn:

> Due to missing perl libraries in newer Debian versions and other problems with sendxmpp I wrote new new tool named txxmpp.c based on the widely available strophe library.  We use it to push commit announcements to the gnupg-devel MUC.  It is a plain and simple tool with all instructions in the help output.  I am not sure whether it fits into the client list but I have not seen a better place, yet.

I put this into a new commit hoping this will do the fix! See:
https://github.com/xsf/xmpp.org/pull/612

### DIFF
--- a/data/clients.json
+++ b/data/clients.json
@@ -636,7 +636,7 @@
         "last_renewed": null,
         "name": "txmmp",
         "platforms": [
-            "Unix",
+            "Unix"
         ],
         "url": "https://git.gnupg.org/cgi-bin/gitweb.cgi?p=gnupg-doc.git;a=blob;f=tools/txxmpp.c"
     },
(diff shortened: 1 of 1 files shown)


data/clients.json
```diff
@@ -631,6 +631,15 @@
         ],
         "url": "https://trillian.im"
     },
+    {
+        "doap": null,
+        "last_renewed": null,
```
SeveFP, line 636: If it does not have a date It will not be listed if I'm not mistaken, since the point of this list is to keep it up to date.
Can you add a date, please?

Thanks in advance @Echolon

SeveFP (changes requested):

data/clients.json
Echolon: Done

PR #852, opened 2020-12-14
Fix months in standards process for 'Deferred'

melvo:
See https://xmpp.org/extensions/xep-0001.html#states-Deferred


Echolon:
@wurstsalat3000 it was fine right?

wurstsalat3000:
Yes!

Echolon (approved):
Thanks for this spot! @melvo

PR #853, opened 2020-12-15
Update members and roles

agnauck:
Update members and roles as per Q3-2020 and board&council elections

### DIFF
--- a/data/clients.json
+++ b/data/clients.json
@@ -636,7 +636,7 @@
         "last_renewed": "2020-12-10T19:00:55",
         "name": "txmmp",
         "platforms": [
-            "Unix"
+            "Linux"
         ],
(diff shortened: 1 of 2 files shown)


data/clients.json
```diff
@@ -636,7 +636,7 @@
         "last_renewed": "2020-12-10T19:00:55",
         "name": "txmmp",
         "platforms": [
-            "Unix"
+            "Linux"
```
Zash, line 639: This looks like something that should not be part of this PR

agnauck: I only added this because the build validation was always failing with linter errors

agnauck: But I can remove when you want to fix the linter errors in a separate PR

PR #857, opened 2020-12-29
Add a dark variant for the main website

linkmauve:
After https://github.com/xsf/xeps/pull/1010 it hurts the eyes to go back to the list, this PR fixes that.

I’ve never managed to get Pelican to build, so this was developed by downloading the rendered HTML file from xmpp.org and modifying it until it looks good, but it doesn’t look like it’ll cause any issue.

Here is how it looks: wayland-screenshot-2020-12-29_17-56-57

### DIFF
--- a/xmpp.org-theme/static/css/app.css
+++ b/xmpp.org-theme/static/css/app.css
@@ -6,3 +6,28 @@
 li p {
     margin-bottom: 0;
 }
+
+@media (prefers-color-scheme: dark) {
+	body {
+		background-color: #161616;
+		color: #ccc;
+	}
+
+	label {
+		color: #aaa;
+	}
+
+	h1, h2, h3, h4, h5, h6 {
+		color: #aaa;
+	}
+
+	table {
+		background-color: #333;
+		border: 1px solid #444;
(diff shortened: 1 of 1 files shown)


Echolon:
@Zash do you think we can merge this?

Zash:
The site using absolute URLs everywhere makes it difficult to try it out, since it uses the published style instead of the new changed style.

So, sure, why not, merge it! Version control makes it enough to revert in case something doesn't work right.

Echolon:
Thank you @linkmauve for your PR!

Echolon:
@linkmauve is it working as expected?

Zash:
Screen Shot 2021-05-21 at 21 19 18

Looks like it did something at least! :partying_face:

Echolon:
Great! How do I provoke this? :smile:

Zash:
I believe browsers mostly follow the OS theme settings these days, so look for appearance or theme settings and see if there's a dark theme. Browser's developer's tools may also have a toggle for debugging.

Echolon:
Yes worked!

PR #858, opened 2021-01-02
Update poezio’s DOAP file.

linkmauve:
poezio’s main branch changed to main, this also synchronises slixmpp and poezio’s renewal time.


Echolon (approved):
Thank you for updating!

PR #868, opened 2021-01-11
Add an @alt to images which didn’t have one

linkmauve:
This can be useful for accessibility tools, or during development to know which image corresponds to that one missing image.


wurstsalat3000:
Fixed in 1842fc7e5f3f167d52fc0f35a40cf8a81b3203e4

PR #880, opened 2021-01-29
Fix missing file extensions

Zash:
Some of the recent blog translations were missing the `.md` file extension, so they were probably invisible to Pelican.


Echolon (approved):
Thank you @Zash

Issue #866, opened 2021-01-10
Move txxmpp into "libraries/tools"


SamWhited:
Please consider replacing the "Libraries" page with "Libraries/Tools" and moving `txxmpp` from the clients page to there.

Alternatively, please consider removing txxmpp entirely because it doesn't have a website or any info and a link to a random C source file with no information isn't especially helpful, even if you know what it is and how to build it.

Echolon:
Hello Werner @ddj9n,

I guess this affects your commit amd software. May you point out or eventually improve regarding the above statement?

Related PR: #849

Echolon:
@SamWhited I agree that we should change the section's title as you suggested. May you have time for a short PR?

dd9jn:
> Please consider replacing the "Libraries" page with "Libraries/Tools" and moving `txxmpp` from the clients page to there.
> 
> Alternatively, please consider removing txxmpp entirely because it doesn't have a website or any info and a link to a random C source file with no information isn't especially helpful, even if you know what it is and how to build it.

Well, there is information right at the top of the source and for such a little tool this should be sufficient.  If you don't think so I could provide a simple web page, just give me a few days.

dd9jn:
I was wrong, there is no info at the top but a --help output with example and such.  Should be easy to find for any hacker.  Usually I put usage info into comments but here I thought a --help would be better.

SamWhited:
> @SamWhited I agree that we should change the section's title as you suggested. May you have time for a short PR?

No, sorry, I don't know how to use this weird website thing they're using.

>  Should be easy to find for any hacker.

This is not easily install-able or usable, let's just not list it. This page isn't here to list every random source file that sends a message that anyone has ever written. It's not for "hackers", it's for random people who think the idea of XMPP sounds interesting to find clients. Even on the libraries page which is more for hackers I'd expect information about what the thing is, not just a link to a source file.

PR #883, opened 2021-01-29
Libaries: Expand to tools

Echolon:
The reason behind this is the txxmpp commit and discussion in #866

@SamWhited


Echolon:
@guusdk & @SamWhited are you fine with this?

SamWhited:
I don't know anything about the website or how it works, but I definitely think if we're going to list tools they shouldn't be mixed in with clients, so sounds good to me.

PR #897, opened 2021-02-19
Add Aparté in clients list

paulfariello:
Add a console client I'm writting  in rust.

https://github.com/paulfariello/aparte/

### DIFF
--- a/data/clients.json
+++ b/data/clients.json
@@ -10,6 +10,17 @@
     },
     {
         "doap": null,
+        "last_renewed": "2021-02-19T09:06:23",
+        "name": "Aparté",
+        "platforms": [
+            "BSD",
+            "Linux",
+            "macOS"
+        ],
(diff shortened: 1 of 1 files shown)


data/clients.json
```diff
@@ -8,6 +8,17 @@
...
         "url": "https://adium.im"
     },
+    {
+        "doap": null,
+        "last_renewed": "2021-02-19T09:06:23",
+        "name": "Aparté",
+        "platforms": [
+            "Linux",
+            "macOS",
+            "*BSD",
```
guusdk, line 18:
```suggestion
            "BSD",
```

Continuous Integration fails. I suspect this is because the platform `*BSD` is not recognized. Try using `BSD` instead.

paulfariello: Thanks, first issue was the trailing comma and second one was the order.

Echolon (approved):
Thanks!

PR #900, opened 2021-02-23
Renew all IgniteRealtime projects

guusdk:
This renews all active projects from the IgniteRealtime community, and removes a couple that are discontinued.


Echolon (approved):
Thanks!

PR #908, opened 2021-03-10
update members as per Q1-2021 elections

agnauck:


Echolon (approved):
Great!

Echolon:
Removed members will be informed?

Echolon:
> > Removed members will be informed?
> 
> I would assume they get a notification from Mailman when I remove them

I know its more work, but I'd recommend to write that personally, also inviting them to come back.

PR #909, opened 2021-03-25
Add bylaws to menu under /about/xsf

Zash:
Because I have to find them in my browser history every single time I'm
looking for them. And this is usually the spot where I look first.

Note: Also a commit that fixes an off-by-one in the menu counter(???)
that presumably unintentionally excluded the comm-team link.


Echolon (approved):
Great @Zash Thanks!

PR #912, opened 2021-04-04
Add txmmp to libraries.json

Echolon:
@dd9jn this is not related, correct? https://github.com/rpavlik/txmpp


Echolon:
@dd9jn please remind to update the timestamp once a year.

Thanks for reaching out to place the entry. Sorry, that it took ages.

PR #913, opened 2021-04-10
XSF people updates

mwild1:
Some small updates to the XSF/people pages:

- Clarified that Seve resigned from the current Board term.
- Update that I'm currently in the Executive Director position.


Echolon (commented):
Thanks!

PR #915, opened 2021-04-22
Add fiscal host policy

SamWhited:
Add an initial draft of the policy for fiscal hosting.

### DIFF
--- a/config/_default/menus.en.toml
+++ b/config/_default/menus.en.toml
@@ -134,6 +134,11 @@
     parent = "community"
     weight = 7
 [[main]]
+    name = "Fiscal Hosting"
+    url = "community/fiscalhost"
+    parent = "community"
+    weight = 8
+[[main]]
     name = "Getting Started"
     url = "/getting-started"
     weight = 5
--- a/content/community/fiscalhost.md
+++ b/content/community/fiscalhost.md
@@ -1,0 +1,25 @@
+---
+Title: Fiscal Host
+Url: community/fiscalhost
+---
+
+The XSF is a U.S. 501(c)(3) organization.
+It sometimes acts as a fiscal host for open source projects related to XMPP.
+
+The XSF may choose to accept organizations at the discretion of the board,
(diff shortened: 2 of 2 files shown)


SamWhited:
What does me being an assignee mean? Is there something I need to do here?

Echolon:
No, I actually means nothing. I just try to keep things a bit organised

Echolon:
But the money will be donated directly to the proct, not via the XSF?

SamWhited:
Sort of. The money is donated to the XSF and earmarked for that particular project. This is called "fiscal sponsorship". Eg. if you don't have the resources to do accounting and create an organization to own your project you can let the XSF take money on your behalf and distribute it to you, this way your donors can take advantage of the XSF's non-profit status and get a tax break and you don't have to figure out how to keep track of donations.

Echolon:
Ah, ok yes is good. I was just wondering if we want to set any limits on that or whether we have to? (Well, it would be nice if people start to donate lot of course!)

mwild1 (changes requested):
Assuming approval by the board, this PR will need some small tweaks now that we are using Hugo. Only to the metadata should be affected, and probably only removing some of the redundant entries is all that's needed. Other pages in the same section should serve as an example of what's needed.

It's okay to postpone any changes until after definite approval of the policy.

SamWhited:
Updated for Hugo. I also took the liberty of adding a menu item under the community menu.

wurstsalat3000 (approved):

SamWhited:
Maybe this should all be under "about" instead?

mwild1 (approved):

mwild1:
Merged as-is. We can still move it if desired. I personally did look for it under "About" first, but I have no strong feelings. It is a service for the community, so that makes equal sense to me.

SamWhited:
Thank you! I am enabling applications in Open Collective now and adding the link to this page!

Echolon (commented):
@SamWhited can you make the HTTP links as a reference behind the words?

PR #921, opened 2021-05-19
Add BombusMod and Jimm Multi

bodqhrohro:


Echolon:
Hi @bodqhrohro,

thank you for your PR!

I have two comments on this:

- What is J2ME?
- The Jimm software is an archived project and hasnt been developed for 8 years. I dont think we should add this to the list. Would you agree?

bodqhrohro:
> or the phones that predate the smartphone era

The smartphone era started way back in late 90s, while J2ME remained 
relevant until circa 2014, when the development of commercial J2ME 
software, like Opera Mini or Gameloft games, finally ceased and only 
enthusiasts kept making new apps.

J2ME is not just a limited-resource platform, but also a universal 
platform which truly implements the Java's motto "write once, run 
everywhere". (This niche was superseded by Web technologies in 10s 
though, when it occasionally turned out that almost every platform has 
some kind of a web browser ;)) J2ME apps run not only on feature phones 
with low-end proprietary operating systems like OSE or Nucleus, but also 
on high-end smartphones, communicators and PDAs powered by Symbian, WM, 
Blackberry OS and Palm OS. It was often considered a first and only 
platform to develop an app to target any mobile devices, as developing 
native apps for every platform is costful, just like technologies like 
Cordova or React Native are used now for the similar purpose.

J2ME apps can be run on modern Android devices as well, in emulators 
like Microemulator or J2ME Loader. Web-based emulators like Kava are 
being developed too, which brings J2ME apps on KaiOS. Some modern 
Chinese feature phones, like Nokia 3310 3G (2017), Mafam M11 (2019), or 
Itel 5630 (2020), still come with a J2ME VM out-of-box.

It's up to you what additions to accept, of course XMPP keeps backwards 
compatibility, thus even the clients that were abandoned way back in 
early 00s still can be used with almost no fuss. Yeah, they lack modern 
XEPs like MAM or File Upload, but many of modern clients lack that as 
well I've seen no notices that the list is intended for actively 
supported software only, but it's okay if this is assumed implicitly.

(Also, I've noticed that [the 
page](https://xmpp.org/software/clients.html) that I came here from 
lacks many of clients present in `clients.json`, even those surely alive 
like Adium, Kaidan or Miranda-NG, probably because it filters them by 
`last_renewed` date, and no one bumped it — should I?)

Echolon:
@bodqhrohro thank you for the extended response!

I would be fine to merge the BombusMod. But the other thing would rather be somthing for an archive, right?

If you want to do the efforts and check if those clients are more or less present, then yes, that would be a great support!

bodqhrohro:
> But the other thing would rather be somthing for an archive, right?

I spotted some other abandoned and archived projects in `libraries.json`, like `jabber.net`, which was superseded by `Sharp.Xmpp`, or `seesmic-as3-xmpp`, which development was stopped in 2009. If the list is not intended for abandoned projects, should those be removed too?

Or should I maybe set an older `last_renewed` date for `Jimm Multi`, to reflect that is wasn't updated for a long time?

Echolon:
Sorry @bodqhrohro - I need to follow the policy. But many thanks for the efforts. Sorry, that I ask you to go ahead :(

If you still would like to support here, we are collaborating on the XMPP Newsletter. There you definitively can contribute! https://xmpp.org/newsletter.html

PR #922, opened 2021-05-21
Update last_renewed for some clients

bodqhrohro:
I used evidence like last commit/push time, or last release date, for proprietary software.

### DIFF
--- a/data/clients.json
+++ b/data/clients.json
@@ -1,7 +1,7 @@
 [
     {
         "doap": null,
-        "last_renewed": "2019-10-20T00:00:00",
+        "last_renewed": "2021-03-13T16:14:00",
         "name": "Adium",
         "platforms": [
             "macOS"
@@ -61,7 +61,7 @@
     },
     {
         "doap": null,
-        "last_renewed": null,
+        "last_renewed": "2021-04-11T14:14:00",
         "name": "BitlBee",
         "platforms": [
             "Linux"
@@ -70,7 +70,7 @@
     },
     {
         "doap": null,
-        "last_renewed": "2020-11-28T00:00:00",
+        "last_renewed": "2021-04-19T00:00:00",
         "name": "blabber.im",
         "platforms": [
             "Android"
@@ -128,7 +128,7 @@
     },
     {
         "doap": null,
-        "last_renewed": "2020-05-13T19:46:58",
+        "last_renewed": "2021-01-18T17:43:00",
         "name": "ChatSecure",
         "platforms": [
             "iOS",
@@ -138,7 +138,7 @@
     },
     {
         "doap": null,
-        "last_renewed": null,
+        "last_renewed": "2021-05-21T14:24:00",
         "name": "Chatty",
         "platforms": [
             "Linux"
@@ -147,7 +147,7 @@
     },
     {
         "doap": null,
-        "last_renewed": null,
+        "last_renewed": "2013-04-25T00:00:00",
         "name": "Coccinella",
         "platforms": [
             "Linux",
@@ -208,14 +208,14 @@
         "name": "eyeCU",
         "platforms": [
             "Linux",
-            "Other",
+            "OS/2",
             "Windows"
         ],
         "url": "https://eyecu.opiums.eu"
     },
     {
         "doap": null,
-        "last_renewed": null,
+        "last_renewed": "2021-03-03T05:30:05",
         "name": "Finch",
         "platforms": [
             "Linux"
@@ -234,7 +234,7 @@
     },
     {
         "doap": null,
-        "last_renewed": null,
+        "last_renewed": "2018-05-10T09:18:00",
         "name": "GNU Freetalk",
         "platforms": [
             "Linux"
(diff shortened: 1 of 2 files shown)


data/clients.json
```diff
@@ -128,7 +128,7 @@
     },
     {
         "doap": null,
-        "last_renewed": "2020-05-13T19:46:58",
+        "last_renewed": "2021-01-18T17:43:00",
         "name": "ChatSecure",
```
Echolon, line 132: Let's not be too enthusiastic, I am following the iOS developments and I know this is not actively being developed anymore (unfortunately). I would expect the developer to update preferably, if he reconsiders to start over.

data/clients.json
```diff
@@ -147,7 +147,7 @@
     },
     {
         "doap": null,
-        "last_renewed": null,
+        "last_renewed": "2013-04-25T00:00:00",
```
Echolon, line 150: The idea is not to simply list all clients. Ideally, the developer update here themselves. Hence, I would not like to update this here.

data/clients.json
```diff
@@ -208,14 +208,14 @@
...
-            "Other",
+            "OS/2",
             "Windows"
         ],
         "url": "https://eyecu.opiums.eu"
     },
     {
         "doap": null,
-        "last_renewed": null,
+        "last_renewed": "2021-03-03T05:30:05",
```
Echolon, line 218: Maybe this is getting a bit emotional, but I don't think we should put efforts in placing Pidgin here. The XMPP implementation is really old and as long as the maintainer comes here, I would refuse to update this.

data/clients.json
Echolon: Sorry, I just see you are setting old time stamps - sorry for that!

data/clients.json
```diff
@@ -252,13 +252,15 @@
     },
     {
         "doap": null,
-        "last_renewed": "2019-07-16T20:02:35",
+        "last_renewed": "2021-01-12T00:00:00",
```
Echolon, line 255: It doesn't even support XMPP anymore. Let's remove it entirely please. It is also know as scammer/bloat software.

data/clients.json
```diff
@@ -321,7 +323,7 @@
     },
     {
         "doap": null,
-        "last_renewed": "2018-08-29T15:00:00",
+        "last_renewed": "2021-05-20T18:25:00",
```
Echolon, line 326: I'm pretty sure it is not developed anymore: https://github.com/jitsi/jitsi/releases/tag/Jitsi-2.10

data/clients.json
```diff
@@ -374,7 +376,7 @@
     },
     {
         "doap": null,
-        "last_renewed": "2019-07-16T20:18:31",
+        "last_renewed": "2021-05-21T00:00:00",
```
Echolon, line 379: I'm having doubt here, too

data/clients.json
```diff
@@ -444,7 +449,7 @@
     },
     {
         "doap": null,
-        "last_renewed": "2019-07-16T20:26:34",
+        "last_renewed": "2021-05-18T07:08:18",
```
Echolon, line 452: I'm having doubt here, too, see my comment on Finch

data/clients.json
```diff
@@ -737,7 +742,7 @@
     },
     {
         "doap": null,
-        "last_renewed": "2019-07-16T20:55:15",
+        "last_renewed": "2021-02-09T00:00:00",
```
Echolon, line 745: Please remove it entirely: https://github.com/zom/Zom-iOS-XMPP
It not even using XMPP anymore.

Echolon:
@bodqhrohro thanks for doing a lot of efforts! I still have some change request. I hope you agree on this!

Echolon (changes requested):
@bodqhrohro thanks for doing a lot of efforts! I still have some change request. I hope you agree on this!

guusdk:
Although I appreciate the effort, the XSF board decided on a policy where only people associated to the projects should be allowed to add/renew the corresponding listing (for the record: I didn't agree with that, but still, it is policy).

Echolon:
Ah okay, then I have to close this unfortunately.

Evtl. it is possible to talk to the relevant developers of at least the developed clients?

data/clients.json
bodqhrohro: > The XMPP implementation is really old

... and still the best among what Pidgin supports out-of-box at all. The implementations for IRC and Bonjour are way more primitive, and the rest of prpls seem to be just a pile of dead code for dead protocols.

If Pidgin is not an XMPP-first client anymore, then what is that, just a framework for great third-party prpls? ;)

> as the maintainer comes here

Should I notify them?

Echolon: @bodqhrohro - I recommend to do the following: 

Reach out the developer of interest and ask to update the timestamp **IF**

- they are still actively developing and maintaining their project / XMPP implementation of their project
- think it is worth to guide new people to their project

Echolon: For all the stale projects we might create a separate archive collection

grimmy:
fwiw, pidgin is aware and okay not being on this list until we get our xmpp implementation to a minimum of the modern xmpp spec.

Echolon:
> fwiw, pidgin is aware and okay not being on this list until we get our xmpp implementation to a minimum of the modern xmpp spec.

Alright - thank you for the feedback! We will not update as long as the developers does it

PR #932, opened 2021-06-11
update members as per Q2-2021 elections

agnauck:


Echolon (approved):
@agnauck feel free to ping me, then I see that imediately!

Welcome @wurstsalat3000 <3 !!11!!

PR #934, opened 2021-06-22
Change link to correct site of CommTeam

Echolon:
From there is also links to the Wiki


SeveFP:
Thank you @Echolon  :)

PR #937, opened 2021-07-01
[newsletter] Add UWPX Release News

COM8:
As I remember, the image has to be treated specially here.
Let me know if I can help you with that.


Echolon (approved):
Thank you @COM8 !

Great work you did this, also great you considered a news yourself!

I will take care on the picture

PR #946, opened 2021-08-03
Add Mellium Dev Communiqué to the July newsletter

SamWhited:
If possible, please add this post to the newsletter. Though I plan to release these in the month after (like the newsletter itself), the content is about last months dev work so I think it's appropriate in this months newsletter.


Echolon (approved):
Exception approved :P

SamWhited:
Thanks! If I actually keep it up this will basically always be how it works, but I think it makes sense since it's a monthly update of the previous month; chances are we can't actually publish that until the month is over :)

PR #950, opened 2021-08-08
Fix link to Isode whitepaper

peteley:
The link on the Myths page to the Isode whitepaper on low-bandwidth
XMPP seems to have changed on their end.


Kev:
LGTM

PR #953, opened 2021-08-15
Switch to Hugo site generator

wurstsalat3000:
This PR migrates xmpp.org to use Hugo for static site generation.

**Here is what I did:**
* switched to Hugo:
    - migrated templates (processing of members.json, {clients,servers,libraries}.json, xeplist.json)
    - added `prepare_xep_list.py`, which downloads and converts xeplist.xml to xeplist.json
    - adapted `lint-list.py` to directory changes
    - added `announcement.json` for displaying Event announcements on xmpp.org
* created a theme similar to the one used by Pelican for the current deployment, but based on Bootstrap 5.1 (and Fontawesome 5.15.3)
* added a simple user CSS (no compilation necessary)
* added redirects from Pelican URL structure to Hugo URL structure: no .html suffix (aliases in content front matter) + some blog/categories redirects done via /deploy/xsf.conf nginx config file)
* updated Makefiles, Dockerfiles, etc:
   - .travis.yml (adapted to Hugo, passing)
   - Dockerfile (adapted to Hugo and updated to Debian Bullseye, works)
   - fabfile.py (adapted to Hugo, untested)
   - Makefile (adapted to Hugo, works)
   - MakefileDocker (adapted to Hugo, works)
   - Vagrantfile (adapted to Hugo and updated to Ubuntu 21.04, works)

Fixes: #611 , #506 , #468

Obsoletes: #788 , #762 , #743 , #621 , #520 , #508 

For Github to be able to show a diff for changes done to .md files in /content, I first restructured the directory (move/rename) in a separate commit, and then applied changes to the files themselves (mostly concerning frontmatter) in another commit. Viewing the `Migrate content` commit should yield a usable diff when viewed from github.com.

PR #960, opened 2021-09-08
create a central redirect for 'current Compliance Suite'

ge0rg:
It makes some sense to have a central link for the "current" Compliance Suite, that will get updated every year.

We have https://xmpp.org/about/compliance-suites.html as a description of the CS, so having https://xmpp.org/about/compliance-suites-current seemed logical to me.


Echolon:
Please don't merge until #953 has been merged

Echolon:
Hi @ge0rg ,

we have been upgrading from Pelican to the Hugo website generator. Could you review this in case it is still relevant?

That would be really great!

ge0rg:
I've updated the commit for the new content directory structure.

mwild1 (approved):

PR #961, opened 2021-09-13
Extensions: Replace Draft status with Stable

linkmauve:
This matches the change in XEPs themselves.


Echolon:
@linkmauve great thanks! We will merge this after #953 has been merged, ok? Then hopefully your other work will be valued aswell

wurstsalat3000:
@linkmauve could you rebase this on current master please?

Relevant files:

* https://github.com/xsf/xmpp.org/blob/master/themes/xmpp.org/layouts/shortcodes/extensions-table.html
* https://github.com/xsf/xmpp.org/blob/master/tools/prepare_xep_list.py

linkmauve:
Here, rebased.  Sorry for the delay!

PR #964, opened 2021-09-18
Extensions: List new and recently changed XEPs

Zash:
Goal is to show new and recently updated XEPs more prominently.

I wanted to make these into three columns (like on [modules.prosody.im](https://modules.prosody.im/)), but I could not get nested shortcodes to work.

image

### DIFF
--- a/content/extensions.md
+++ b/content/extensions.md
@@ -5,7 +5,24 @@ aliases:
     - "/extensions/index.html"
     - "/protocols/xmpp-extensions/"
 ---
-The core specifications for XMPP are developed at the Internet Engineering Task Force (IETF) - see [RFC 6120](https://datatracker.ietf.org/doc/rfc6120/), [RFC 7590](https://datatracker.ietf.org/doc/rfc7590/), [RFC 6121](https://datatracker.ietf.org/doc/rfc6121/), and [RFC 7622](https://datatracker.ietf.org/doc/rfc7622/) (along with a WebSocket binding defined in [RFC 7395](https://datatracker.ietf.org/doc/rfc7395/)).
+
+## IETF specifications
+
+-   [RFC 6120](https://datatracker.ietf.org/doc/rfc6120/) XMPP Core
+-   [RFC 6121](https://datatracker.ietf.org/doc/rfc6121/) XMPP IM
+-   [RFC 7622](https://datatracker.ietf.org/doc/rfc7622/) XMPP Address Format
+-   [RFC 7590](https://datatracker.ietf.org/doc/rfc7590/) Use of TLS in XMPP
+-   [RFC 7395](https://datatracker.ietf.org/doc/rfc7395/) XMPP over WebSockets
+
+## New XEPs
+
+{{< new-xeps >}}
+
+## Recently updated
+
+{{< updated-xeps >}}
+
+## The whole list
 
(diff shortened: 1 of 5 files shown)


wurstsalat3000:
I found two approaches for a column based layout:

* put the whole thing into one shortcode
* make use of something like this:

**extensions.md**
```
{{< row >}}

{{< col header="IETF specifications" >}}
-   [RFC 6120](https://datatracker.ietf.org/doc/rfc6120/) XMPP Core
-   [RFC 6121](https://datatracker.ietf.org/doc/rfc6121/) XMPP IM
-   [RFC 7622](https://datatracker.ietf.org/doc/rfc7622/) XMPP Address Format
-   [RFC 7590](https://datatracker.ietf.org/doc/rfc7590/) Use of TLS in XMPP
-   [RFC 7395](https://datatracker.ietf.org/doc/rfc7395/) XMPP over WebSockets
{{< /col >}}

{{< col header="New XEPs" content="shortcode" >}}
{{< new-xeps >}}
{{< /col >}}

{{< col header="Recently updated" content="shortcode" >}}
{{< updated-xeps >}}
{{< /col >}}

{{< /row >}}
```

**col.html**
```
<div class="col">
    {{ if .Get "header" }}
    <h2>{{ .Get "header" }}</h2>
    {{ end }}
    {{ if eq ( .Get "content") "shortcode" }}
        {{ .Inner }}
    {{ else }}
        {{ .Inner | markdownify }}
    {{ end }}
</div>
```

**row.html**
```
<div class="row">
    {{ .Inner }}
</div>
```

Zash:
Thanks, making some progress now!

image

(obs: `<table>`s for layout :rofl: )

Zash:
Found [how to make columns in this version of Bootstrap](https://getbootstrap.com/docs/5.1/layout/columns/)

image

Zash:
Narrow view:
image

Where did the bootstrap things that did smart things go?

Zash:
EVEN NARROWER!

image

wurstsalat3000:
Column distribution and stacking works if you apply breakpoints to the *col classes. For example:

* three columns of 4 = 12 on xl (≥1200px)
* two columns of 6 = 12 on md (≥768px)
* one column of 12 on < md

```
<div class="col-xl-4 col-md-6">
```

Zash:
Oh those still exist? Didn't see them in the bootstrap docs.

Zash:
> ```
> <div class="col-xl-4 col-md-6">
> ```

This seems to work, can't say I understand why.

wurstsalat3000:
Sure: https://getbootstrap.com/docs/5.0/layout/grid/#grid-options

You opted for `col-md-4`, which looks pretty squeezed when closing to the md-breakpoint:

image

With an alternative approach you could also go the other way around and let the row's class do all the work:

Using three plain `<div class="col">` and a `<div class="row row-cols-1 row-cols-md-2 row-cols-xl-3">` parent, you'll get the same result as with a `<div class="row">` combined with three `<div class="col-xl-4 col-md-6">` children.

Zash:
> use the force, @Zash

Okay, `git push -f`

wurstsalat3000 (approved):

Echolon:
This is great work! Thanks!

PR #969, opened 2021-09-21
Use subresource integrity on stylesheets

SamWhited:
This PR adds subresource integrity checks to all stylesheets and minimization to the stylesheets that we control.
This acts as a cache key, validating that the served resource is in fact what it's expected to be and can be used to invalidate long lived caches if the resource changes.


wurstsalat3000 (approved):
Works fine, thank you!

SamWhited:
I feel like I've asked this before, but this got assigned to me, but I'm also the original author; is there something I should do to it?

PR #970, opened 2021-09-22
Deduplicate use of subtitle in theme

SamWhited:
Previously the subtitle was used several times in the theme. This mixes
content and theming so instead pull the subtitle from the config so that
if it ever changes the change is reflected everywhere.


wurstsalat3000 (approved):
Good idea! Works fine.

Issue #414, opened 2018-03-16
RFCs are missing and maybe not updated for actual


Neustradamus:
The RFCs list is not complete and maybe not updated for actual (last 2019-06-21):
- https://xmpp.org/rfcs/

Status problems:
- RFC 3920
- RFC 3921
- RFC 3922
- RFC 3923 (no change)
- RFC 4622
- RFC 4854 (no change)
- RFC 5122
- RFC 6120
- RFC 6121 (no change)
- RFC 6122

New to old not on the website:
- RFC 8600
- RFC 8284
- RFC 8266
- RFC 8048
- RFC 7712
- RFC 7702
- RFC 7700
- RFC 7622
- RFC 7590
- RFC 7573
- RFC 7572
- RFC 7395
- RFC 7340
- RFC 7259
- RFC 7248
- RFC 7247
- RFC 7165
- RFC 7081
- RFC 5437
- RFC 4979

wurstsalat3000:
Wouldn't it be nicer to have an overview page at xmpp.org/rfcs, which would list and describe relevant RFCs, and link to datatracker.ietf.org instead? This would surely minimize maintenance effort.

Issue #242, opened 2017-01-09
RFC 3920 not marked as deprecated


SamWhited:
Over the last few days I've seen people link to the version of RFC 3920 on the website (https://xmpp.org/rfcs/rfc3920.html) and reference it multiple times. When I pointed out that it was deprecated they were suprised.

Maybe we should add a deprecation notice to the copy of RFC 3920 on the website?

guusdk:
I tried introducing a banner or a landing page mentioning the deprecation, but the RFC's don't appear to be part of the website code at all. I'm not sure how to modify these.

For what it's worth: on http://xmpp.org/extensions/index.html we don't link to the copy that's on our own server, but to something on https://datatracker.ietf.org instead.

horazont:
Can we simply re-direct to the HTML version from the IETF? I personally find it much easier to read anyways.

guusdk:
I think the RFC's are in this repository: https://github.com/xsf/rfcs

Those are XMLs though, not HTML. @Kev, can you help out here?

Flowdalic:
> Those are XMLs though, not HTML. @Kev, can you help out here?

The HTML versions are generated using `xml2rfc --html`.

Neustradamus:
Any news?

wurstsalat3000:
Hey guys, it’s been three years ;)

Linking to the corresponding RFCs on https://datatracker.ietf.org/ sounds sensible to me.

Neustradamus:
@wurstsalat3000: Thanks for this recall :)

Please look again: https://github.com/xsf/xmpp.org/issues/414 etc.

Echolon:
Actually, where exactly does it need to be deprecated?

Here it is obsolete: https://datatracker.ietf.org/doc/rfc3920/

wurstsalat3000:
@Echolon it's about the one hosted on xmpp.org, not the one at https://datatracker.ietf.org/doc/rfc3920/

Echolon:
> @Echolon it's about the one hosted on xmpp.org, not the one at https://datatracker.ietf.org/doc/rfc3920/

but isn't the status at IETF not the "true" one?

Neustradamus:
All files from: https://xmpp.org/rfcs/.

Example:
- https://xmpp.org/rfcs/rfc3920.html
- https://xmpp.org/rfcs/rfc3920.txt
- https://xmpp.org/rfcs/rfc3920.xml

All IETF files are good.

[...]

PR #975, opened 2021-09-23
Add RFC overview page

wurstsalat3000:
This adds an overview page listing XMPP RFCs.

The overview page is automatically generated via `prepare_rfc_list.py`, which pulls relevant RFC reference files from ietf.org. These files are then processed and summarized into a single `rfc_list.json`, which can be displayed by Hugo. For each RFC the following infos are displayed:

* title
* author(s)
* year
* abstract
* link to RFC at datatracker.ietf.org
* if available: link to RFC `html` / `txt` / `xml` hosted at xmpp.org/rfcs/

The new overview page is available at xmpp.org/rfcs and is meant as a replacement for https://xmpp.org/rfcs/index.shtml .

Each RFC details block is linkable via an anchor, meaning xmpp.org/rfcs#6120 will bring you directly to the RFC's details.

Improves the situation around #414

rfc-overview


Echolon:
Will this also be close then? https://github.com/xsf/xmpp.org/issues/242

wurstsalat3000:
@Echolon not necessarily.

This PR would certainly help to improve the situation around #242. But it won't update the RFC 3920 document hosted on xmpp.org.

#224 is not affected in any way by this PR.

Neustradamus:
@wurstsalat3000: Good job! :)

Can you look to solve XEP problems?
- https://github.com/xsf/xeps/issues/966

PR #977, opened 2021-09-25
Add reference to Modern XMPP project

Echolon:


wurstsalat3000:
I'd replace `would like to reference to the independent` with `would like to suggest the independent`.

PR #985, opened 2021-10-06
Fix typo.

mdosch:
xmpp-dns v0.2.0 brings support for xmpps-server, not xmpp-server (which was included from the beginning).


Echolon:
Ok, my bad - thanks!

PR #986, opened 2021-10-08
Render clients list using processed DOAP files

### LINKED ISSUE #405: Logos in client list
It would be great to support additional fields in the client / server / (library?) list:

 * logo
 * download (for mobile clients this would be the respective platform store)
 * maybe also a description (which should have an appropriate length limit)?

I'd like to do the start:
 * Bruno: https://op-co.de/tmp/bruno-64.png
    * https://play.google.com/store/apps/details?id=org.yaxim.bruno
 * Conversations:  https://raw.githubusercontent.com/siacs/Conversations/master/art/ic_launcher.svg
    * https://play.google.com/store/apps/details?id=eu.siacs.conversations
 * yaxim: https://op-co.de/tmp/yak-front-grass.svg
    * https://play.google.com/store/apps/details?id=org.yaxim.androidclient

wurstsalat3000:
This adds a new rendering of the clients list to xmpp.org/software/clients.

Client data is gathered by using DOAP files linked in `data/clients.json`.

Clients are sorted by platform and displayed in 'cards'.

The resulting `clients_list_doap.json` is generated automatically by the tooling supplied by this PR. Logos are downloaded (and resized, if necessary) by the tooling as well.

Closes #405 

image

### DIFF
--- a/tools/prepare_software_list.py
+++ b/tools/prepare_software_list.py
@@ -1,0 +1,306 @@
+'''
+Download / prepare / process XMPP DOAP files for the software list
+Requires: Pillow, python-slugify
+'''
+from typing import Optional
+from typing import Union
+from typing import cast
+
+from datetime import datetime
+from datetime import timedelta
+from pathlib import Path
+import json
+import os
+import shutil
+from urllib.parse import urlparse
+
+from defusedxml.ElementTree import parse
+from defusedxml.ElementTree import ParseError
+import requests
+from PIL import Image
+from PIL import UnidentifiedImageError
+from PIL.Image import Resampling
+from slugify import slugify
+
+DOWNLOAD_PATH = Path('downloads')
+DATA_PATH = Path('data')
+STATIC_PATH = Path('static')
+LOGOS_PATH = STATIC_PATH / 'images' / 'clients'
+
+DOAP_NS = 'http://usefulinc.com/ns/doap#'
+SCHEMA_NS = 'https://schema.org/'
+RDF_RESOURCE = '{http://www.w3.org/1999/02/22-rdf-syntax-ns#}resource'
+DOAP_NAME = f'.//{{{DOAP_NS}}}name'
+DOAP_SHORTDESC = f'.//{{{DOAP_NS}}}shortdesc'
+DOAP_OS = f'.//{{{DOAP_NS}}}os'
+DOAP_LOGO = f'.//{{{SCHEMA_NS}}}logo'
+
+PLATFORMS: list[str] = [
+    'Android',
+    'iOS',
+    'Windows',
+    'macOS',
+    'Linux',
+]
+
+ENTRY_LIFETIME = timedelta(days=365)
+
+
+def initialize_directory(path: Path) -> None:
+    '''
+    Remove path (if it exists) and containing files, then recreate path
+    '''
+    if path.exists() and path.is_dir():
+        shutil.rmtree(path)
+        os.mkdir(path)
+    else:
+        os.mkdir(path)
+
+
+def download_file(url: str, path: Path) -> bool:
+    '''
+    Downloads file from url and stores it in /downloads/path
+    returns success
+    '''
+    file_request = requests.get(url, stream=True, timeout=5)
+    if not 200 >= file_request.status_code < 400:
+        print('Error while trying to download from', url)
+        return False
+
+    with open(DOWNLOAD_PATH / path, 'wb') as data_file:
+        max_size = 1024 * 1024 * 10  # 10 MiB
+        size = 0
+        for chunk in file_request.iter_content(chunk_size=8192):
+            data_file.write(chunk)
+            size += len(chunk)
+            if size > max_size:
+                file_request.close()
+                print('File size exceeds 10 MiB:', path)
+                return False
+    return True
+
+
+def check_renewal(name: str, renewed: Optional[str]) -> bool:
+    '''
+    Check if 'last_renewed' entry is not older than 365 days
+    '''
+    if renewed is None:
+        return False
+    try:
+        renewed = datetime.strptime(renewed, "%Y-%m-%dT%H:%M:%S")
+    except ValueError:
+        print('Failed to parse timestamp for', name)
+        return False
+    now = datetime.utcnow()
+    if now - renewed > ENTRY_LIFETIME:
+        print('Client entry expired for', name)
+        return False
+    return True
+
+
+def parse_doap_infos(doap_file: str
+                     ) -> Optional[dict[str, Union[str, list[str], None]]]:
+    '''
+    Parse DOAP file and return infos
+    '''
+    try:
+        doap = parse(
+            DOWNLOAD_PATH / f'doap_files/{doap_file}.doap')
+    except (FileNotFoundError, ParseError):
+        return None
+
+    info: dict[str, Union[str, list[str], None]] = {}
+
+    info['name'] = None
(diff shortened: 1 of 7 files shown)


Echolon:
Hi @linkmauve 

still no luck to find a bit of time? I think this is an awesome feature I would really like to see published.

But I respect if you haven't

tools/prepare_software_list.py
```diff
@@ -0,0 +1,290 @@
...
+
+from datetime import datetime
+from datetime import timedelta
+from pathlib import Path
+import json
+import os
+import shutil
+import sys
+from urllib.parse import urlparse
+from xml.etree import ElementTree as ET
```
horazont, line 17: https://docs.python.org/3/library/xml.html

> **Warning**
> 
> The XML modules are not secure against erroneous or maliciously constructed data. If you need to parse untrusted or unauthenticated data see the [XML vulnerabilities](https://docs.python.org/3/library/xml.html#xml-vulnerabilities) and [The defusedxml Package](https://docs.python.org/3/library/xml.html#defusedxml-package) sections.

As we are downloading XML from random servers, we have to ensure that it's not going to blow up, especially now that we build the website on our servers again and not in GitHub's or Travis' CI :).

tools/prepare_software_list.py
```diff
@@ -0,0 +1,290 @@
...
+from PIL.Image import Resampling
+from slugify import slugify
+
+DOWNLOAD_PATH = Path('downloads')
+DATA_PATH = Path('data')
+STATIC_PATH = Path('static')
+LOGOS_PATH = STATIC_PATH / 'images' / 'clients'
+
+XSF_CLIENTS_LIST_URL = 'https://raw.githubusercontent.com/xsf/xmpp.org/master/' \
+    'data/clients.json'
```
horazont, line 30: Shouldn't this be available locally?

tools/prepare_software_list.py
```diff
@@ -0,0 +1,290 @@
...
+    Downloads file from url and stores it in /downloads/path
+    returns success
+    '''
+    file_request = requests.get(url)
+    if not 200 >= file_request.status_code < 400:
+        print('Error while trying to download from', url)
+        return False
+
+    with open(DOWNLOAD_PATH / path, 'wb') as file_data:
+        file_data.write(file_request.content)
```
horazont, line 73: Please use [`iter_content`](https://docs.python-requests.org/en/latest/api/#requests.Response.iter_content) and enforce an absolute limit on the file size. We don't want a misconfigured or malicious server fill our disk or RAM (`.content` buffers the entire response in-memory).

A reasonable limit would be 10 MiB IMO.

tools/prepare_software_list.py
```diff
@@ -0,0 +1,290 @@
...
+
+def check_file_size(file_path: Path) -> None:
+    '''
+    Check if file size is greater than 300 KiB and if so, resize image
+    '''
+    try:
+        file_size = os.path.getsize(file_path)
+    except OSError as error:
+        print('An error occurred while trying to open logo:', error)
+    if file_size > 300000:
```
horazont, line 138: Nit:

```python
if file_size <= 300000:
    return
```

saves a level of indentation.

tools/prepare_software_list.py
```diff
@@ -0,0 +1,290 @@
...
+        try:
+            with Image.open(file_path) as img:
```
horazont, line 140: I'm not sure I'm happy with PIL processing untrusted data, but I guess that is ok if we ensure that pillow and the backing libraries are up-to-date.

tools/prepare_software_list.py
```diff
@@ -0,0 +1,290 @@
...
+    image_url = urlparse(uri)
+    _, extension = os.path.splitext(image_url.path)
+    file_name = f'{client_name}{extension}'
+    success = download_file(
+        uri,
+        Path(file_name))
+    if not success:
+        return None
+
+    if extension[1:] in ('png', 'jpg'):
```
horazont, line 166: What about `jpeg`, `bmp`, ...?

Should we enforce an allowlist on file extensions?

tools/prepare_software_list.py
```diff
@@ -0,0 +1,290 @@
...
+
+
+def prepare_clients_list() -> None:
+    '''
+    Download and prepare clients data
+    '''
+    success = download_file(
+        XSF_CLIENTS_LIST_URL, Path('xsf_clients_list.json'))
+    if not success:
+        sys.exit()
```
horazont, line 183: [If I read `sys.exit`](https://docs.python.org/3/library/sys.html#sys.exit) correctly, this exits without any message and without error code. Is that truly what you want?

I would've expected an exception raised here, or an error printed and exit with non-zero exit code, because this effectively makes this thing not do what it's supposed to do.

tools/prepare_software_list.py
```diff
@@ -0,0 +1,290 @@
...
+
+
+def parse_doap_infos(doap_file: str
+                     ) -> Optional[dict[str, Union[str, list[str], None]]]:
+    '''Parse DOAP file and return infos'''
+    try:
+        doap = ET.parse(
+            DOWNLOAD_PATH / f'doap_files/{doap_file}.doap')
+    except FileNotFoundError:
+        return None
```
horazont, line 102: This doesn't guard against any parser errors. Any invalid XML of a single individual client would, as I read it, break the entire listing.

Please protect against syntactically invalid XML.

horazont (changes requested):
Sorry for the brevity, reviewing this in a 10 minute slot before work.

Please address the security concerns before this can get merged, thanks!

tools/prepare_software_list.py
wurstsalat3000: It is indeed, thanks.

tools/prepare_software_list.py
wurstsalat3000: switched to `defusedxml` for parsing

tools/prepare_software_list.py
wurstsalat3000: Good point, thanks! 10 MiB it is.

tools/prepare_software_list.py
wurstsalat3000: My intent was to not touch SVG files. I refactored this a bit. But I don't think we need an explicit allowlist, do we?

tools/prepare_software_list.py
wurstsalat3000: Since `clients.json` is always available locally, this code isn't necessary anymore.

tools/prepare_software_list.py
wurstsalat3000: I'm catching `defusedxml.ElementTree.ParseError` now.

tools/prepare_software_list.py
wurstsalat3000: At least `Pillow` is installed via `pip install --upgrade`, and I didn't pin any version in `requirements.txt`.

wurstsalat3000:
> Sorry for the brevity, reviewing this in a 10 minute slot before work.
> 
> Please address the security concerns before this can get merged, thanks!

@horazont thank you very much for this review! I hope I addressed all points of your feedback?

tools/prepare_software_list.py
wurstsalat3000: Website builds are now available via Github Actions. This should relax the situation.

Echolon:
😍😍😍😍😍😍

Neustradamus:
@wurstsalat3000: Thanks!

PR #988, opened 2021-10-19
Add default blog post parameters

SamWhited:
This makes it easier for users to know what options they can fill out
when starting a new blog post using "hugo new". Other options that should be filled out by users writing blog posts can be added later.


wurstsalat3000:
No, it does not affect the build.

Thanks!

PR #989, opened 2021-10-20
Config: allow unsafe rendering of raw HTML

ge0rg:
This is needed for some pages (e.g. old Summits) that use `<table>` and
other raw elements.

Affected page example:

https://xmpp.org/2010/08/xmpp-summit-4/ (scroll down to "**Room shares and room reservations for XSF members.**")

Current rendering:

image

Expected / fixed rendering:

image

### DIFF
--- a/config/_default/config.toml
+++ b/config/_default/config.toml
@@ -24,3 +24,6 @@ enableEmoji = true
 
 [permalinks]
   blog = "/:year/:month/:title/"
+
+[markup.goldmark.renderer]
+unsafe = true


ge0rg:
I *think* the implications are only relevant if we allow untrusted users to submit markdown source, which I hope we don't.

There is a workaround to only allow raw HTML in pages when surrounded by a shortcode, see https://discourse.gohugo.io/t/enable-unsafe-true-for-a-single-page/24415/4 but I think that won't change the security properties in any useful way.

wurstsalat3000:
I'll close this for now. Please let me know if there is anything I missed.

PR #993, opened 2021-10-22
Add office hours signup form

SamWhited:
I've been thinking that it might be nice to advertise the XMPP office hours a bit more on the community page of the website. Furthermore, it would be nice if the comm team were actually alerted when someone signs up. To do the second part I have been experimenting with a netlify form that emails us with the results (which we can then put on the schedule). Embedding it in the page looked nice and accomplished both goals in one. Thoughts?

Screenshot of a form for submitting office hours signup requests

### DIFF
--- a/config/_default/menus.en.toml
+++ b/config/_default/menus.en.toml
@@ -144,6 +144,11 @@
     parent = "community"
     weight = 8
 [[main]]
+    name = "Office Hours"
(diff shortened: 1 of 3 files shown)


SamWhited:
This got merged, but never deployed; I think this happened last time too, maybe something is wrong with the pipeline?

PR #998, opened 2021-10-30
Translate clients.md to German

Echolon:
I don't know how to deal with the meta header yet

### DIFF
--- a/content/software/client.de.md
+++ b/content/software/client.de.md
@@ -1,0 +1,14 @@
+---
+title: XMPP Clients
+aliases:
+    - "/software/clients.html"
+---
+
+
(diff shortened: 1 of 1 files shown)


wurstsalat3000:
Please note that there is no obvious way for accessing translated _pages_ yet. Only _posts_ feature a meta data header with links to access translates _posts_. Markdown files need to have the same base name in order to be linked. There is a typo in `client.de.md`, which should be `clients.de.md`.

Apart from that: I thought we would offer translations for _posts_ only, not for _pages_ ?

Echolon:
> Please note that there is no obvious way for accessing translated _pages_ yet. Only _posts_ feature a meta data header with links to access translated _posts_. Markdown files need to have the same base name in order to be linked. There is a typo in `client.de.md`, which should be `clients.de.md`.
> 
> Apart from that: I thought we would offer translations for _posts_ only, not for _pages_ ?

I understood selected pages like this would have been also fine? But if not, then close.

Issue #152, opened 2016-03-14
Security headers missing


petmue:
Hi!  Please consider introducing some of the following security headers:

> Strict-Transport-Security, Content-Security-Policy, Public-Key-Pins, X-Frame-Options, X-XSS-Protection, X-Content-Type-Options

Currently you make use of none, compare https://securityheaders.io/?q=https%3A%2F%2Fxmpp.org%2F

aerth:
hey just saying setting `Strict Transport Security` _incorrectly_ will mess things up for lots of people until they clear their cache, so be **careful**.

linkmauve:
aerth, you are probably thinking about HPKP instead, HSTS doesn’t cause any risk as long as you keep valid certificates.

SamWhited:
I mix those up every time too; I agree though, if you don't have a big team and lots of automation and alerting then setting `Public-Key-Pins` is just asking for trouble.

horazont:
Excluding ``Public-Key-Pins`` and maybe ``Referrer-Policy``, I think the headers are not really relevant to a static website without confidential information.

intosi:
With our use of LetsEncrypt certs, I'm not comfortable using HPKP currently.

salarsk:
9368437163

guusdk:
Is there anything left to be done?

Echolon:
The author of is issue did delete his account. Are there any security issues know regarding this issue? Otherwise I would close the issue.

@guusdk 
@horazont

guusdk:
This is well outside of my area of expertise.

SamWhited:
`X-XSS-Protection` doesn't really matter as long as `Content-Security-Policy` is set (which it probably should be). `X-Content-Type-Options` probably can't hurt either, but isn't as necessary these days as long as we are deliberately setting `Content-Type` correctly (which is the more important thing to do).

PR #1001, opened 2021-11-04
Add content security and referrer policy to website

SamWhited:
I've put together a default policy that allows the office hours form, loading most resources from the `xmpp.org` domain, and disallows a few more dangerous resources that we don't use.

This also adds a referrer policy (no referrer sent on HTTPS/HTTP downgrade, otherwise just send the full thing since there are no "private" pages on the site and we may want others to know that we linked to them in a blog post or what not.

Other very minor changes were required to remove eg. unsafe JavaScript that would be blocked by the new policy.

Updates #152

### DIFF
--- a/deploy/xsf.conf
+++ b/deploy/xsf.conf
@@ -26,6 +26,9 @@ server {
         try_files $uri $uri/ $uri.html =404;
     }
 
+    add_header Content-Security-Policy "default-src 'self' xmpp.org; img-src *; connect-src 'none'; object-src 'none'; child-src xmpp-office-hours.netlify.app; frame-src xmpp-office-hours.netlify.app; worker-src 'none'; frame-ancestors 'none'; form-action 'self' xmpp-office-hours.netlify.app; upgrade-insecure-requests; block-all-mixed-content";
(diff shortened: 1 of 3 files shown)


Echolon:
ahh, it does not close #152 ?

SamWhited:
It does by implementing some of the most important headers. The commit message should be correct.

Echolon:
@ge0rg could you also take a look?

Echolon:
I cannot review this, is there anyone volunteering?

linkmauve:
Besides netlify.app which I know nothing about (but the restrictions can be strengthened later, it’s already better to allow just this one than to allow every single website ever), and the 404 page seemingly never being used, LGTM!

Echolon:
@Zash or @mwild1 any final concerns about this?

Ppjet6:
I'm not entirely sure the SRI tag is useful here but it doesn't hurt. LGTM too otherwise.

mwild1 (approved):

Echolon:
Many thanks to the recent reviews!

@wurstsalat3000 if you dont see any further reuqired edits feel free to finally merge.

wurstsalat3000:
Thanks!

PR #1002, opened 2021-11-05
Update clients.json

rockliffe:
Renew AstraChat.

### DIFF
--- a/data/clients.json
+++ b/data/clients.json
@@ -30,7 +30,7 @@
     },
     {
         "doap": null,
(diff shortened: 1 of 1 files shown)


Echolon (approved):
Thanks!

Echolon:
@rockliffe another hint here: we are rebuilding our  clients pages website rendering: #986

These will make further use of the DOAP files and also render the clients logo. If you host something like it can be referemced and will also be shown for your software.

https://xmpp.org/extensions/xep-0453.html

Several clients in the list have such files public available to look at how it can look like.

PR #1004, opened 2021-11-13
New xmpp client

geobra:
Add Shmoose, a xmpp client primary for the mobile operating system Sailfish OS. It's also runable on Linux.

### DIFF
--- a/data/clients.json
+++ b/data/clients.json
@@ -754,5 +754,5 @@
             "iOS"
         ],
         "url": "https://zom.im/"
-    }   
+    }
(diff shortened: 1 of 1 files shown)


Echolon (changes requested):
Hi @geobra,

many thanks! One minor detail: Can you put it in the right order in the file?

That would be great! Looking forward!

geobra:
Sorted it alphabetic.

Echolon (approved):
Many thanks @geobra !

PR #1006, opened 2021-11-14
Update officehours.md

Echolon:
@SamWhited


SamWhited:
I think it was fine before personally. It's obvious that it's the XMPP office hours because it's on a site dedicated to XMPP and the contractions don't matter, but whatever you want to do 🤷

Echolon:
I know, but we always announce it as XMPP Office hours so I preferred to have it aligned.

Let @wurstsalat3000 decide :D

PR #1010, opened 2021-11-26
update roles per annual board and council election

agnauck:
update roles per annual board and council election


Echolon (approved):
Congratulations to everyone!

Echolon:
@mwild1 can you run a new deploy if you find some time?

PR #1013, opened 2021-11-26
update board and council section

agnauck:
update board and council section after our annual election.
see also #1013


agnauck:
pushed an initial commit.
Will push more updates once I receive bios for JC and Travis

PR #1019, opened 2021-11-29
Add IEEE paper to publications

Echolon:


wurstsalat3000:
Thanks! I added this in e1f94291930a705812722178fa4cb814a10a6bca

jcbrand:
There are also these (not sure whether we should be adding every research paper however):

https://ieeexplore.ieee.org/document/8320956
https://ieeexplore.ieee.org/document/8394998
https://ieeexplore.ieee.org/document/9153556

Echolon:
@jcbrand yes great! I would say that we just update it and if it gets overwhelmingly too much we should consider a different solution. but linking for now is good and also to give people something to eat for the moment :-)

wurstsalat3000:
I went forward and added [papers.json](https://github.com/xsf/xmpp.org/blob/master/data/papers.json) with a custom rendering.

Echolon:
Really awesome 😍 @wurstsalat3000

PR #1024, opened 2021-11-30
Fix Shmoose platform order

mwild1:
The site failed to build because platforms must be in alphabetical order:

```
python3 /var/tmp/src/xmpp.org/tools/lint-list.py clients.json
ERROR: entry 'Shmoose': platform order must be: 'Linux', 'Sailfish OS' (platforms must be ordered alphabetically)
Found 1 severe violations. Please fix them.
```

### DIFF
--- a/data/clients.json
+++ b/data/clients.json
@@ -542,8 +542,8 @@
         "last_renewed": null,
         "name": "Shmoose",
         "platforms": [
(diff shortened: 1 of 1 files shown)


Echolon (approved):
Ah thanks

PR #1038, opened 2021-12-21
Update announcement.json

Echolon:
@Flowdalic @wurstsalat3000


wurstsalat3000:
This announcement sounds rather formal, I think. How about something like:

"Google Summer of Code 2022 - dive into XMPP" or "Google Summer of Code 2022 - {insert motivation}"

PR #1048, opened 2022-01-23
Renew Conversations

Neustradamus:
Renew Conversations


wurstsalat3000:
See #1055

Neustradamus:
@wurstsalat3000: I think that first PR is better ;)

@iNPUTmice has done a duplicate of mine, do not mix.

PR #1060, opened 2022-01-25
Update GSoC_2022_Logo.png with transparency

Echolon:


wurstsalat3000:
That image weights 4.1 Megabytes! Please reduce the file size.

PR #1067, opened 2022-02-04
Use a better logo for Linux on the homepage

linkmauve:
The previous logo looked pretty dull, especially in the dark theme (see the screenshot), this one is colorful and thus better.

wayland-screenshot-2022-02-04_12-17-09

I traced simple geometry behind the existing logo and ran [svgcleaner](https://crates.io/crates/svgcleaner) on it, in order to not increase its size too much.


Echolon:
I hope everything will go fine and just merge! YOLO

PR #1068, opened 2022-02-05
Add canonical link

Zash:
Supposedly helps SEO or something, especially when you might have the
same site served on both www.example.com and example.com
Also nice if you ever downloaded a page to know where it came from


Echolon (approved):
I trust @Zash and will just merge

PR #1074, opened 2022-02-14
Renew poezio & slixmpp data

mathieui:
A bit overdue but fine.


Echolon (approved):
Thanks - I know the practice is not the best: I recommend to maybe create an annual entry to your calendar to keep it alive! but yeah, I know...

PR #1076, opened 2022-02-18
Added Stanza as Client Side Library

Kalmuraee:


data/libraries.json
```diff
@@ -803,5 +803,14 @@
...
         "url": "http://ivan.vucica.net"
+    },
+     {
+        "doap": null,
+        "last_renewed": null,
+        "name": "Stanza",
+        "platforms": [
+            "TypeScript"
+        ],
+        "url": "http://stanza,io"
```
Echolon, line 814: I think you should replace the , with a .

Echolon:
@Kalmuraee thanks for the PR!

Be reminded that we require project owners to update the timestamp annually (otherwise it will not be shown on the website).

In general I recommend to provide a DOAP file too as we will render clients, server etc. based on these files in the future.

https://xmpp.org/extensions/xep-0453.html

@legastero can you confirm this PR

data/libraries.json
Kalmuraee: Resolved, Seems I'lI need another glasses :)

wurstsalat3000:
Since there is no feedback from the maintainer, I'm going to close this for now. Please re-open if this changes in the future.

PR #1081, opened 2022-02-22
Add Github action for software list linting

wurstsalat3000:
This adds software list linting via Github Actions.

An example run can be seen here: https://github.com/wurstsalat3000/xmpp.org/runs/5291903236?check_suite_focus=true


Echolon:
Ahh many thanks

Echolon:
But we still need to fix the file, right?

PR #1099, opened 2022-03-07
Renew MatriX XMPP libraries

agnauck:


Echolon (approved):
You are also invited regarding adding a doap file.

PR #1111, opened 2022-03-22
data: renew Mellium library registration

SamWhited:
Renewing. Left the day the same so it doesn't creep forward and make me adjust my calendar reminder by a week every year.

### DIFF
--- a/data/libraries.json
+++ b/data/libraries.json
@@ -390,7 +390,7 @@
     },
     {
         "doap": null,
-        "last_renewed": "2021-03-29T00:00:00",
+        "last_renewed": "2022-03-29T00:00:00",
         "name": "Mellium",
         "platforms": [
             "Go"


SamWhited:
We can just wait for a few days to merge if it will refuse to build or something too. It does seem like it should be "more than a few weeks in the future or more than the previous date + 1 year" or something though and not strictly "in the future from when the PR was submitted".

guusdk:
I think it's 24 hours now: https://github.com/xsf/xmpp.org/blob/master/tools/lint-list.py#L85-L93

Also, why doesn't that workflow run for this PR? It surely used to. I distinctly remember it catching issues with PRs.

wurstsalat3000:
Should work now for upcoming PRs.

wurstsalat3000:
Raising this limit to 1 week seems sensible to me. What do you think?

guusdk:
I don't think we need to be very protective here. The objective is only to get rid of clear attempts to manipulate the process. I'd be fine with four weeks. It helps people renewing approximately once a year without to much hassle, while any 'abuse' is still minimal. 

Also very much not a hill that I'm willing to die on. Whatever works.

PR #1130, opened 2022-05-13
Update monal timestamp

tmolitor-stud-tu:
Why do we have to periodically update an arbitrary timestamp to make the website show our client?

Do I have to write a github workflow to automatically create pull requests every few weeks?


Echolon:
@tmolitor-stud-tu Thanks for your PR. The idea is to present clients to users that actually experience some development and are not legacy or unmaintained.

The renewal is only required once a year.

PR #1139, opened 2022-05-28
Mention Kik Messenger

wasamasa:
See https://www.forbes.com/sites/thomasbrewster/2017/09/20/kik-slow-to-delete-child-abuse-profiles-despite-promise/ for the 15 million active users figure.


guusdk:
@Echolon that only applies to projects to be added to the software listings on the site. This PR adds to the 'uses' section - basically our bragging rights.

As such, I wouldn't be to concerned about the definition of a user, and simply go with the higher value. :)

wasamasa:
Is there anything I should change about the PR?

Echolon:
> Is there anything I should change about the PR?

Maybe we can go with the 300 mio. from the article

PR #1146, opened 2022-06-26
German translation newsletter april and may

wh0nix:
Hey guys,

here is the german translation for the newsletters april and may.


wurstsalat3000:
Thanks a lot!

PR #1155, opened 2022-08-01
Update libraries.json

ammarheidari:


Echolon (approved):
Thanks!

Echolon:
@ammarheidari consider this for libraries as well: https://xmpp.org/extensions/xep-0453.html

We do this for clients for the website rendering: https://xmpp.org/software/clients/
But we might plan to expand this for servers and libraries as well.

Sample: https://github.com/conversejs/converse.js/blob/master/conversejs.doap

PR #1161, opened 2022-08-11
Allow projects to put logo URL in JSON or DOAP

SamWhited:
Currently only DOAP projects can have a nice logo, and those of us without are being left behind. This patch adds a logo field to the JSON so that the site doesn't have as many ugly projects with generic logos and everyone can have a pretty logo.

### DIFF
--- a/data/libraries.json
+++ b/data/libraries.json
@@ -395,7 +395,8 @@
         "platforms": [
             "Go"
         ],
-        "url": "https://mellium.im/code"
+        "url": "https://mellium.im/code",
+        "logo": "https://mellium.im/mellium_icon.png"
     },
     {
         "doap": null,
--- a/tools/lint_list.py
+++ b/tools/lint_list.py
@@ -13,12 +13,15 @@ from datetime import datetime
 from datetime import timedelta
 
 
-VALID_ENTRY_KEYS = {
+REQUIRED_ENTRY_KEYS = {
     "platforms",
     "name",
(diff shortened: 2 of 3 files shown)


tools/lint_list.py
```diff
@@ -13,13 +13,16 @@
 from datetime import timedelta
 
 
-VALID_ENTRY_KEYS = {
+REQUIRED_ENTRY_KEYS = {
```
SamWhited, line 16: I split these out because otherwise every single field has to have a duplicate `logo: None` that will never be used. It doesn't really matter though. If we do want to keep it we should consider what other keys can be moved, but I was trying to keep this patch as minimal as possible so I didn't include them here.

data/libraries.json
```diff
@@ -395,7 +395,8 @@
         "platforms": [
             "Go"
         ],
-        "url": "https://mellium.im/code"
+        "url": "https://mellium.im/code",
+        "logo": "https://mellium.im/mellium_icon.png"
```
SamWhited, line 399: I couldn't get the site working to tell if this is actually working, but it does build so that's a good sign?

wurstsalat3000:
There is now a basic DOAP file for Mellium hosted at xmpp.org. You can find it at https://github.com/xsf/xmpp.org/blob/master/static/doap/mellium.doap

The rendered version (including a logo) is available at https://xmpp.org/software/libraries/

PR #1162, opened 2022-08-12
anchor tags to XMPP apps

josamontiel:
Added anchor tags to xmpp apps with blank targets to allow for loading in new tab, it improves the user experience because it allows the user to bounce back and forth between the client page and the 'Getting Started' page

### DIFF
--- a/content/getting-started.md
+++ b/content/getting-started.md
@@ -12,25 +12,25 @@ There are [many, many XMPP clients](/software/clients) for you to choose from. T
 
 ### Desktop
 
-  * [Beagle IM](https://beagle.im/) (macOS)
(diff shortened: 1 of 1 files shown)


josamontiel:
Not sure if it was suggested already but thought I'd give my 2 cents on that. 

Looking forward to y'alls response

wurstsalat3000:
Thanks for your contribution! I added this with a slightly different solution to avoid inline html, see e3d5c0c484b9897b51b2f6f8c6a61cf3cbc89e61

josamontiel:
No problem, been loving XMPP, so glad I got to help in any way!

PR #1170, opened 2022-09-07
Add Cheogram Android to clients list

singpolyma:


Echolon (changes requested):
Great!

Can you specify the supported platforms?

singpolyma:
It's Android. Does that not come from the doap? I copied the JSON Conversations was using and it doesn't seem to need the platforms in the JSON either so I assumed it came from the DOAP.

PR #1184, opened 2022-10-16
Update slixmpp and poezio listings

Ppjet6:


guusdk:
18th of October? Someone is _cheating_! ;)

PR #1195, opened 2022-11-03
Update clients.json

georgehazan:
- mark Miranda IM as dead;
- update DOAP information about Miranda NG

### DIFF
--- a/data/clients.json
+++ b/data/clients.json
@@ -389,14 +389,16 @@
         "platforms": [
             "Windows"
         ],
-        "url": "http://miranda-im.org"
+        "url": null
     },
     {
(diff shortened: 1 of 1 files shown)


wurstsalat3000:
Thanks! I changed both entries in c325cc3ca1ccbbc413e0ef6b78c48f8d5e7c70d1

"platform" and "url" entries are not needed if a DOAP file is present.

Neustradamus:
@wurstsalat3000: There is a problem, your changes have broken Miranda IM entry, it is now missing from the file (old projects are always in files but not showed on the website).
There is only Miranda NG in the file.

georgehazan:
@Neustradamus is that a problem? Miranda IM is dead

PR #1200, opened 2022-11-08
Add jsxc logo

sirdrakeistda:
self explaining


wurstsalat3000:
@sirdrakeistda a proper solution would be to persue JSXC's authors to host a DOAP file in _their_ repository, where they can add a logo.

wurstsalat3000:
See https://github.com/jsxc/jsxc/issues/1115

Echolon:
Closing, comeback when the time is right @sualko

PR #1206, opened 2022-11-15
Add DOAP files for JaXMPP library

woj-tek:


wurstsalat3000:
Thanks a lot!

wurstsalat3000:
@woj-tek as I have you here:

Some of Tigase's software listed in {clients,libraries,servers}.doap show discrepancies between what's given as "name" in for example clients.json, and what's given as "name" element in their respecive DOAP files.

Here is an example:

from clients.json:
```
"name": "SiskinIM by Tigase, Inc.",
```

from siskin-im.doap:
```
<name>Siskin IM</name>
```

I noticed this while debugging xmpp.org's script, which adds compliance data to the listing for example at https://xmpp.org/software/clients/siskinim-by-tigase-inc/

For Siskin IM there should be a nice set of badges, since the `compliancer`  shows:
```
{"name":"Siskin IM","type":"client","badges":{"av":"advanced","core":"advanced","im":"core","mobile":"advanced"}}
```
But it isn't shown, because the name doesn't match. I observed this for JaXMPP and other Tigase software as well. It would be nice if you could unify those names.

woj-tek:
@wurstsalat3000 created https://github.com/xsf/xmpp.org/pull/1208

Though, considering - it could be convenient to add something along the lines of author/project/company name field?

PR #1208, opened 2022-11-16
Unify Tigase naming

woj-tek:


wurstsalat3000:
> No need to add the "website" field though, since it's already covered by the DOAP files. We try to minimize duplication of entries (client names are an exception here).

@woj-tek sorry, I meant "No need to add the "url" field [...]", which was meant for `clients.json` etc.

wurstsalat3000:
Alright, I found other packages with naming discrepancies as well, and fixed those in 9dfe349c38bd38eb3407e21907579d63df40aa6a together with Tigase's packages. I added Martin in ecc9173475afa70ee986aadd8a5cb3da0600d6e2

Thanks again!

PR #1213, opened 2022-11-25
remove outdated years

agnauck:
remove the years, so we don't have to update them annually


Echolon (approved):
I wish we would got the resources to be detailed on our belongings, be don't :(

@mwild1 concerns?

PR #1215, opened 2022-12-01
Renew Prosody

Zash:
That time of the year again.

Weren't we going to replace this procedure with DOAP?


Echolon (approved):
Yes, TBA

Until then, thanks for reminding!

PR #1221, opened 2022-12-13
update memberlist Q4-2022 election

agnauck:
update memberlist after Q4-2022 election results


singpolyma:
My jid should be `singpolyma@singpolyma.net`

Echolon (approved):
Welcome!

PR #1230, opened 2023-01-09
Update public server listing links

mwild1:
The link to IM Observatory was broken, as that service is now retired. I've replaced it by a link to the new providers.xmpp.net.

I've removed the link to compliance.conversations.im. That data is factored into ratings on providers.xmpp.net and, at the time of writing, the service has an uncertain future.


Echolon:
Hi  @iNPUTmice,

any options that someone could take over your service?

PR #1252, opened 2023-02-27
Add GSoC 2023 promo files (EN, DE)

Echolon:


wurstsalat3000 (commented):
The German version needs to be unified in regard to formal/informal addressing (Sie/Du).

PR #1255, opened 2023-03-01
Fix and rename Service Providers link

Fishbowler:
The Service Providers link (in Community and the right side of some pages) linked to a now-missing link on xmpp.work

I have:
* Fixed the link to point at the root of https://xmpp.work
* Updated the name of the link to more accurately reflect the purpose of xmpp.work

### DIFF
--- a/config/_default/menus.toml
+++ b/config/_default/menus.toml
@@ -119,8 +119,8 @@
     parent = "community"
     weight = 6
 [[main]]
-    name = "Service Providers"
-    url = "https://xmpp.work/all-listings"
+    name = "Jobs using XMPP"
+    url = "https://xmpp.work/"
     parent = "community"
     weight = 7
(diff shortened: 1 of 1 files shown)


wurstsalat3000:
Thanks!

Fishbowler:
Finding the wording on this tricky.

It's Jobs. It's XMPP.

Jobs with XMPP - could be implied to be with XSF?
Jobs using XMPP - could infer "XMPP operator" is some kind of job
XMPP Jobs - could be both

Anything I've summoned that's more accurate is longer, which has its own problems

wurstsalat3000:
Right, it's not easy to find a perfectly fitting title for this menu entry. I think a general "Jobs with XMPP" is best here. It's not "Jobs with the XSF" :)

PR #1267, opened 2023-04-21
/software/: Use words to describe compliance

WhyNotHugo:
The existing approach used a custom font to render some icons as the only way of describing compliance level. For clients that don't load or render custom remote fonts, this information is invisible. For clients that do render fonts, the meaning of these icons is a mystery.

Additionally, for users with screen readers or similar assistive technologies, these icons are also meaningless (especially since they have no "alt").

This patch adds the actual words to describe the compliance level, so that the information can be easily consumed by humans.


WhyNotHugo:
Screenshots without and with remote fonts enabled:

image

image

cal0pteryx:
Thanks a lot! I took your PR for inspiration, which resulted in c9d3d13b2f1202fb755f9c645e1d67154978cab0

PR #1269, opened 2023-04-21
Explicitly use lua5.2

WhyNotHugo:
The binary "lua" might point to a different version. Given that this specific version is indicated as a dependency, make sure the matching version is used.

In particular, this avoids issues on distributions which make 5.1 the default (since some modules are missing for it).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,7 @@ ENV DEBIAN_FRONTEND noninteractive
 RUN apt-get update && apt-get dist-upgrade -y && apt-get autoremove -y && apt-get clean
 
 # Install dependencies.
-RUN apt-get install -y hugo lua5.2 lua-expat python3 python3-pip
+RUN apt-get install -y hugo lua5.4 lua-expat python3 python3-pip
 
 # Base URL for Hugo website builds
 ARG BASEURL=https://xmpp.org/
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ You need to have the following dependencies installed:
 
 * Hugo
 * Python 3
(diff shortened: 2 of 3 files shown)


cal0pteryx:
But this fails the GitHub build.

Echolon:
@WhyNotHugo can you maybe check what's broken?

cal0pteryx:
GitHub Action actually uses Lua 5.4.4 (latest) at the moment. Versions from Lua 5.2 to 5.4 seem to work. We should add a version fix. But this doesn't prevent systems with Lua 5.1 to run the tool chain. To prevent this, we would have to add a version check.

WhyNotHugo:
Right, 5.3 works too. We can either pin 5.4 in the code, or install 5.2 on CI. But I've no idea how to change the lua version in GitHub's environment.

WhyNotHugo:
Huh, it doesn't look like GH uses 5.4 either. I have no way of reproducing this locally, so not sure what to do. I can leave it as-is and just see if I can change my default lua >= 5.2; this might be complicating things too much.

cal0pteryx:
Yes, GitHub uses Lua 5.4 in this Action. Looking at the logs you see "Lua version detected: 5.4". It's just that the version GH provides is called by `lua`, and not by `lua{version}` (i.e. no version alias). That's why I suggested to add a version check in `prepare_compliance.py`.

cal0pteryx:
Should be solved by c1681c225b6e05d3aac43810e976e712c88f77e7 and 431f284a0465004fed0017f4b614424986b636aa

Thanks for your PR!

PR #1298, opened 2023-09-10
Update software.json: Add status cecker

Echolon:
Add status checker

### DIFF
--- a/data/software.json
+++ b/data/software.json
@@ -1954,7 +1954,7 @@
             "tool"
         ],
         "doap": "/hosted-doap/xmpp-account-exporter.doap",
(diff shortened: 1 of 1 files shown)


data/software.json
```diff
@@ -1954,10 +1954,19 @@
             "tool"
         ],
         "doap": "/hosted-doap/xmpp-account-exporter.doap",
-        "name": "XMPP account exporter",
+        "name": "XMPP Account Exporter",
```
cal0pteryx, line 1957: This has to be the same name as the one listed in the tool's DOAP file. Both the DOAP file _and_ the website use lower case writing ;)

PR #1303, opened 2023-09-26
Update software entry of Monal IM

tmolitor-stud-tu:


Echolon (approved):

cal0pteryx:
Both URL and Platform values are duplicated from the DOAP file and not needed in software.json. I reverted these in 9f627e2350f4b675e8bc9681d531c43357c34446

PR #1304, opened 2023-10-03
extensions-table.js: make all XEPs visible when slash/find key pressed

mwild1:
We don't currently have any in-page facility for searching across XEPs (e.g. by title). That's fine, as browsers typically offer this functionality. However currently to perform such a search, you have to tick all the checkboxes manually before performing the search.

This small addition to the code will automatically enable all checkboxes, thus revealing all XEPs, when the user presses the "/" key, which triggers a quick find-in-page functionality in at least Firefox.

Known issues:

- Not all browsers search when "/" is pressed (and catching Ctrl+F is not easy)
- It's not discoverable (this is a feature and a bug - I am a fan of quiet enhancements, but there may be people/environments for which an inline search would be preferable)

These issues should not degrade anyone's experience compared to what we have today, but it will be a small improvement for some people (including me).

### DIFF
--- a/themes/xmpp.org/static/js/extensions-table.js
+++ b/themes/xmpp.org/static/js/extensions-table.js
@@ -19,4 +19,14 @@ window.addEventListener("load", function() {
   jsSupport.forEach(function(elem) {
     elem.classList.remove("jsSupport");
   });
+
+  document.addEventListener("keypress", function (event) {
+    if (event.code != "Slash" && event.code != "Find") {
+      return;
+    }
+    checkboxes.forEach(function (checkbox) {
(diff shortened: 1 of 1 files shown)


cal0pteryx:
I just added an inline search bar in b22f3a1
Does this work for you or should this approach be implemented, too?

cal0pteryx:
@mwild1 ?

mwild1:
The search bar is great! I think for me to be truly happy though, either we merge this in addition, or adapt it to focus the new search input box on keypress (I don't have strong feelings about whether it's okay to override default '/' behaviour).

PR #1307, opened 2023-10-08
Update software _index.md

Echolon:
I would love to see it like this still


cal0pteryx:
Could you explain why?

Echolon:
> Could you explain why?

Yes, because noone will find and use it like this. The list is very long.

cal0pteryx:
You moved the "Legacy Software" list above the DOAP software listing. Legacy entries should be displayed further down, not right at the top.

Echolon:
> You moved the "Legacy Software" list above the DOAP software listing. Legacy entries should be displayed further down, not right at the top.

Oh okay. So what I want is this up:

Screenshot_20231024_194718

PR #1324, opened 2023-11-09
Update instant-messaging.md: Add MOKA

Echolon:
And some general updates

@iNPUTmice any concerns?


iNPUTmice:
I have no more insights into the project other than the information that is publicly available.

PR #1331, opened 2023-12-16
extensions-table: check deferred XEPs

Flowdalic:
Deferred XEPs are just as relevant as the XEPs of other XEP-states that are checked by default. For example, XEP-0447 (experimental) depends on XEP-0103 and XEP-0104 (both deferred).

I don't think we should present a XEP and not its dependencies per default.

### DIFF
--- a/themes/xmpp.org/layouts/shortcodes/extensions-table.html
+++ b/themes/xmpp.org/layouts/shortcodes/extensions-table.html
@@ -26,7 +26,7 @@
                         <label class="form-check-label" for="Experimental">Experimental</label>
                     </div>
                     <div class="form-check form-check-inline">
-                        <input class="form-check-input" type="checkbox" id="Deferred" name="Deferred" />
+                        <input class="form-check-input" type="checkbox" id="Deferred" name="Deferred" checked="checked" />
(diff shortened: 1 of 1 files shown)


Echolon:
@Flowdalic can you check for the pipeline again?

Flowdalic:
> @Flowdalic can you check for the pipeline again?

Done

cal0pteryx:
@Echolon this is no but, not at all.

I asked in the XSF chat if this would be desirable, but I received mixed feedback. There doesn't seem to be consensus about this.

Flowdalic:
In that case I'd like to run this by one of our elected bodies. Either council or board, whatever body feels responsible. I'd go with council first, as I believe the technical body is more suited to decide how we present your technical specifications, but ultimately I don't really care which body decides if we should show deferred XEPs by default or not.

cal0pteryx:
There has been [some discussion](https://logs.xmpp.org/council/2024-03-05#2024-03-05-330671644dfcc97f) about whether or not to show Deferred XEPs by default. Council finds it's not their decision to make, but the general sentiment was leaning towards showing them by default.

There have also been some suggestions on improvements for the table/additional filters/flags, which I'll consider implementing.

PR #1355, opened 2024-03-10
instant-messaging: use official documentation of WhatsApp

fossdd:
Facebook (WhatsApp) officially documented that their applications use a protocol based on XMPP to communicate.

> Partner users’ devices connect to the WhatsApp server using WhatsApp’s proprietary‬ ‭XML protocol (based on XMPP), as described further below.‬


cal0pteryx:
Thanks!

PR #1357, opened 2024-03-15
Extensions: Add hints for experimental/deferred XEPs

cal0pteryx:
image
image

### DIFF
--- a/themes/xmpp.org/assets/css/style.css
+++ b/themes/xmpp.org/assets/css/style.css
@@ -322,6 +322,11 @@ table.sortable th:not(.sorttable_sorted):not(.sorttable_sorted_reverse):not(.sor
     content: " \25B4\25BE";
     color: grey;
 }
+
+button[name="show-xep-implementations"] {
+    min-width: 2.5em;
+}
+
 .extensions-block-listing li {
     font-family: "Exo 2", sans-serif !important;
     overflow: hidden;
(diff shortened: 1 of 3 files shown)


cal0pteryx:
@mwild1 would this be a sensible thing to do?

mwild1:
11:06:22 MattJ> I like it, I think!
11:06:57 MattJ> When I mentioned it, I was thinking about notices at the header of the XEP rendering itself, rather than the listing
11:07:18 MattJ> On the listing, I suspect that it's XEPs with both these warnings that we should hide by default
11:07:45 MattJ> No implementations + no activity == practically dead
11:08:11 MattJ> Implementations + no activity == we need to advance it
11:08:21 MattJ> No implementations + activity == we need implementations

|                                   | Recent activity | No recent activity |
|-----------------------------|----------------------|-------------------------|
| **Has implementations** | 👷 Implement it / 🚀 Advance it    | 🚀 Advance it |
| **No implementations**   | 👷 Implement it! | 🧹 Dormant, hide it by default (if experimental or deferred) |

cal0pteryx:
Thanks! I went ahead and tried to match these cases:

Roughly 120 XEPs are dormant (and would be hidden by default):

image

I'm not sure if this should rather be a checkbox in a new line below the other boxes, plus an explanation label. Note: If this box is unchecked, both Deferred and Experimental XEPs matching the "Dormant" criterion would be hidden, regardless of the "Deferred"/"Experimental" checkbox state. I could add an additonal counter showing how many dormant XEPs are hidden.

The four stages are displayed with corresponding tooltips:

image

image

PR #1374, opened 2024-04-24
Remove Zom

butyoutried:
They switched from XMPP to Matrix a long time ago.

https://github.com/zom/Zom-Android-XMPP
https://github.com/zom/Zom-iOS-XMPP


cal0pteryx:
Thanks!

PR #1381, opened 2024-04-30
Add aTalk client (with a DOAP file)

butyoutried:
DOAP file for aTalk has just been added: https://github.com/cmeng-git/atalk-android/pull/214


cal0pteryx:
Thanks

PR #1384, opened 2024-05-21
data/software.json: specify platforms of Smack

Flowdalic:


cal0pteryx:
Hi @Flowdalic ! Platforms for Smack are listed in its DOAP file directly, see https://github.com/xsf/xmpp.org/blob/master/static/hosted-doap/smack.doap#L10

You can add as many `<os>` tags as you like.

Once there is a DOAP file for a project, the listing in software.json shouldn't contain any more info than the project's name.

Even better would be if the project adopted (and extended) the DOAP file hosted at this repo :)

Flowdalic:
@cal0pteryx thanks for your reply.

Would deleting static/hosted-doap/smack.doap also be an option to achieve what I want to achieve (i.e., that Smack shows up when either 'Android', 'Linux', 'Windows', or 'macOS' is selected on the webpage)?

cal0pteryx:
Only projects with a DOAP file are listed in the filterable software overview. DOAP-less projects are listed in the legacy table at the very bottom of that page.

Flowdalic:
@cal0pteryx Thanks again for your reply. I've updated the doap file also.

Flowdalic:
Hmm, not sure why the CI is failing

PR #1387, opened 2024-05-25
Remove discontinued Google Talk from Overview page

butyoutried:


Echolon:
Hi,

many thanks for the clean-up.

Would it be okay to rather make a historic section below? maybe with some years range?

25 may 2024 12:35:03 butyoutried ***@***.***>:

Neustradamus:
I think it is better to have "such as Google Talk" without a link.

Echolon:
> I think it is better to have "such as Google Talk" without a link.

Can we rather use: https://en.wikipedia.org/wiki/Google_Talk

Neustradamus:
@cal0pteryx: Can you add Google Talk in the cited page?
- https://xmpp.org/uses/instant-messaging/

Note: For me jitsi has stopped the XMPP service a long time ago and there was not the specified number.

butyoutried:
> Hi, many thanks for the clean-up. Would it be okay to rather make a historic section below? maybe with some years range?

Hi, @Echolon. I'm sorry for missing your message :(

PR #1389, opened 2024-06-03
Clarify new moderation approach in operators channel

mwild1:
A few weeks ago we made the decision to switch operators@ to require explicit membership to be able to participate. The channel documentation on the website was not updated at the time to reflect this change, this commit fixes that.

### DIFF
--- a/content/community/operators-rules.md
+++ b/content/community/operators-rules.md
@@ -98,15 +98,11 @@ If the behaviour continues despite those measures or if the transgression is suf
 
 You may notice an occupant of the channel called `authbot`. It is a moderation assistant bot that automatically grants channel membership to users it identifies as server operators. It performs a [contact addresses (XEP-0157)](https://xmpp.org/extensions/xep-0157.html) query on servers and looks for a match with any of the returned XMPP addresses.
 
-We recommend that all public servers publish contact addresses through this mechanism, but it is not a requirement, nor is it required to join the channel from your listed contact addresses.
-
 ### How do I get channel membership?
 
 Membership is automatically granted by authbot when the criteria are met.
 
-If you have alternative proof that you are a server operator or a member of the XMPP Standards Foundation, you may request membership manually from one of the channel admins. Please don't ask for membership in any other circumstances - it won't be given.
-
-Finally, note that channel membership does not imply any kind of immunity from the rules, or endorsement of a user. Membership will be granted and revoked by channel moderators as they need, irrespective of whether a user is a server operator or not.
+Also, please note that channel membership does not imply any kind of immunity from the rules, or endorsement of a user.
 
 ### I want to discuss something that is considered off-topic in this channel, where can I go?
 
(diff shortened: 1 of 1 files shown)


content/community/operators-rules.md
```diff
@@ -91,7 +102,7 @@ We recommend that all public servers publish contact addresses through this mech
 
 ### How do I get channel membership?
 
-First, don't fret about it. Membership is generally not required to participate in the chat, it is only used as an aid for moderation purposes when necessary.
+Membership is automatically granted by authbot when the criteria are met.
 
 If you have alternative proof that you are a server operator or a member of the XMPP Standards Foundation, you may request membership manually from one of the channel admins. Please don't ask for membership in any other circumstances - it won't be given.
```
horazont, line 107: Do we want to keep this?

("you may request membership manually from one of the channel admins")

Seems contradictory to "Participation rights are managed automatically"

mwild1: I'm not sure. Specifically I don't want a flood of manual verification requests, and I'm not sure what manual verification would look like anyway. The only valid case I can think of is if someone *does* publish a contact address via '157, but prefer to join the MUC from a different address (which I think is a valid use case).

I'm fine with continuing to not grant participation rights to people who don't publish contact addresses.

horazont: I would like to remove it.

Otherwise it may create a lot of effort for the moderation team for exactly that reason ("what does manual verification look like").

horazont (approved):

guusdk (commented):
I think it would be more clear if the section "How do I get channel membership" defines the mechanism that is used (which is now described under the 'Who is authbot' FAQ section). 

I propose to re-order the text from both sections like this:

```
### Who is authbot?

You may notice an occupant of the channel called `authbot`. It is a moderation assistant bot that automatically grants channel membership to users, as described in the 'How do I get channel membership' section of this document.
```

```
### How do I get channel membership?

Membership is automatically granted by authbot when the criteria are met. It attempts to identify room participants as server operators. It performs a [contact addresses (XEP-0157)](https://xmpp.org/extensions/xep-0157.html) query on servers and looks for a match with any of the returned XMPP addresses.
```

guusdk:
Oh, this was already merged. Nevermind then, my suggested change isn't something I feel strongly about.

Echolon:
I think the restrictions are a bit harsh.

What about users? We had situation where the operators support chat was not available anymore.

What about interested potential operators asking for help?

Echolon:
Shouldn't his go through Board first?

horazont:
> What about users? We had situation where the operators support chat was not available anymore.

operators@ is not for users. User support has happened there in the past, but that was strictly speaking off-topic.

> What about interested potential operators asking for help?

There are other venues for that, such as xmpp:servers@joinjabber.org.

> Shouldn't his go through Board first? 

no idea (:

mwild1:
> I think the restrictions are a bit harsh.

Sure, and we delayed this action for a long time. But the channel was not fit for purpose. Semi-closed operator discussion venues are very common for other open networks, including email and telephony. A high-signal, low-noise channel for operators is extremely beneficial for the network. It doesn't have to be the same place that provides general support - as noted in previous comments, such venues already exist for that.

Echolon:
Okay I see

PR #1402, opened 2024-06-20
Change DOAP file for aTalk to a maintained one

butyoutried:
https://github.com/cmeng-git/atalk-android/issues/197#issuecomment-2087926469


cal0pteryx:
Thanks! It would probably be best to remove the old file now from the aTalk repository, to avoid confusion.

PR #1404, opened 2024-06-22
Update "Other Software" list

butyoutried:
* Set proper names for XMPPFramework and XMPPHP.
* Add a website for net::XMPP.
* Fix platforms (supported OSes) for some clients, namely mention BSD.
* Replace unmaintained [Ubiety](https://github.com/ubiety/xmpp) library with it's successor, [Ubiety XMPP Core](https://github.com/ubiety/Ubiety.Xmpp.Core) library of the same author.
* Clean up the list. Coversant SoapBox Communicator/Server, Jabber Stream Objects, Jabbim (PyQt XMPP client, not to be confused with Jabbim XMPP provider), Jerry Messenger, Kwickserver, oajabber, Quiet Internet Pager, Talkonaut: completely gone, no website, no repository, no nothing. IM Observatory: [discontinued](https://xmpp.net/directory.php) since 2022. IM+: no longer offers XMPP support, according to [official website](https://plus.im).
* Add [BombusMod](https://github.com/BombusMod/BombusMod), [xmppc](https://codeberg.org/Anoxinon_e.V./xmppc), [Jackline](https://github.com/hannesm/jackline), [xmpp](https://git.sr.ht/~ft/xmpp) (for Plan 9), [Vacuum-IM](https://github.com/vacuum-im/vacuum-im), [Kontalk](https://www.kontalk.org), [JabberCat](https://jabbercat.org), [CoyIM](https://coy.im) clients.

### DIFF
--- a/data/software.json
+++ b/data/software.json
@@ -1261,7 +1261,7 @@
         "platforms": [
             "Perl"
         ],
-        "url": null
+        "url": "https://metacpan.org/pod/Net::XMPP"
     },
     {
         "categories": [
@@ -2142,7 +2142,7 @@
             "library"
         ],
         "doap": null,
-        "name": "xmppframework",
+        "name": "XMPPFramework",
         "platforms": [
             "Objective C"
         ],
@@ -2153,7 +2153,7 @@
             "library"
         ],
         "doap": null,
-        "name": "xmpphp",
+        "name": "XMPPHP",
         "platforms": [
             "PHP"
         ],
--- a/data/software.json
+++ b/data/software.json
@@ -228,6 +228,17 @@
             "client"
         ],
         "doap": null,
+        "name": "BombusMod",
+        "platforms": [
+            "Android"
+        ],
+        "url": "https://github.com/BombusMod/BombusMod"
+    },
+    {
+        "categories": [
+            "client"
+        ],
+        "doap": null,
         "name": "Boogie Chat",
         "platforms": [
             "iOS"
@@ -384,6 +395,19 @@
         "name": "Converse.js",
         "platforms": [],
         "url": null
+    },
+    {
+        "categories": [
+            "client"
+        ],
+        "doap": null,
+        "name": "CoyIM",
+        "platforms": [
+            "Linux",
+            "macOS",
+            "Windows"
+        ],
+        "url": "https://coy.im"
     },
     {
         "categories": [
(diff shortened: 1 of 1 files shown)


data/software.json
```diff
@@ -1994,6 +1953,17 @@
         ],
         "url": "https://www.xabber.com"
     },
+    {
+        "categories": [
+            "client"
+        ],
+        "doap": null,
+        "name": "xmpp",
```
Echolon, line 1961: Does this make sense? shouldn't it be "xmpp for plan 9"

data/software.json
```diff
@@ -2020,7 +1990,9 @@
         ],
         "doap": null,
         "name": "XMPP Status Checker",
-        "platforms": [],
+        "platforms": [
+            "Browser"
```
Echolon, line 1994: I wonder if we need a 'project' category

data/software.json
```diff
@@ -954,18 +937,6 @@
         "platforms": [],
         "url": null
     },
-    {
-        "categories": [
-            "server"
-        ],
-        "doap": null,
-        "name": "Jerry Messenger",
```
Echolon, line 962: I understand the preference to delete unmaintained projects. However, I wonder if we should keep the for archive reasons. XMPP has a history, I would not like to just delete it.

Echolon (commented):
Many thanks for the efforts.

Maybe you can also ask the new projects if they want to come up with a DOAP file?

data/software.json
butyoutried: Well, the program itself is literally just called “xmpp”.

data/software.json
butyoutried: Instead of "platforms: Browser"?

butyoutried:
> Maybe you can also ask the new projects if they want to come up with a DOAP file?

Unfortunately, I don't have time for this. But Vacuum-IM, Kontalk, JabberCat are unmaintained anyway, and almost all the others are barely maintained right now.

data/software.json
cal0pteryx: It's the project's name, so that's fine.

data/software.json
cal0pteryx: @Echolon what do you mean by that? And: is it in the scope of this PR?

data/software.json
cal0pteryx: I don't see a reason in keeping dead projects on the list. What would be the benefit? There is an internet archive for things which don't exist anymore. I don't think this list's purpose is to archive _every_ XMPP software there was in history. Instead, it should list _available_ software.

data/software.json
cal0pteryx: "Browser" is fine for this project.

cal0pteryx:
@butyoutried thanks for your efforts!

PR #1416, opened 2024-07-07
The Italian transleted files have been updated

nicfab:
I updated the file related to the Italian translation of the Newsletter for May and June 2024.


cal0pteryx:
@nicfab I'm sorry but I can't merge this. If you look at the difference here: https://github.com/xsf/xmpp.org/pull/1416/files you'll see that these are not the changes you intended to make, right?

There are two options:

1) You need to synchronize your fork before editing any file, like shown in this screenshot:

image

2) Make your edits by clicking on the "Edit this page" link at the bottom of each article on xmpp.org, like shown here:
image

No. 2) is probably easier.

PR #1435, opened 2024-09-20
Add Prav Android client

bkmgit:


data/software.json
```diff
@@ -387,6 +387,17 @@
         "platforms": [],
         "url": null
     },
+    {
+        "categories": [
+            "client"
+        ],
+        "doap": "https://codeberg.org/prav/prav/raw/branch/master/prav.doap",
```
cal0pteryx, line 394: This file does not exist.

If you provide a DOAP file, `platforms` and `url` won't need to be filled.

cal0pteryx:
Please re-open if you plan to do changes.

data/software.json
bkmgit: File has been created. Made a new merge request https://github.com/xsf/xmpp.org/pull/1450

PR #1436, opened 2024-09-26
add `Whixp` library to the list of softwares

vsevex:


guusdk:
Thanks for taking the time to add your library! While doing so, you changed every line in the file. I'm guessing that an editor formatted to file or something. Can you please try again, without doing that?

PR #1437, opened 2024-09-26
add `Whixp` library to the list of softwares

vsevex:


guusdk:
Thanks for this. As you may have noticed, the validation for this file is fairly strict. Your change is currently being rejected because of this:

> ERROR: entry 'Whixp': should be placed in front of 'Z-XMPP' (all entries must be ordered alphabetically by case-folded name)

Can you please apply the desired change(s)?

guusdk (approved):

cal0pteryx:
@vsevex some changes are required for this to display correctly on xmpp.org

1. DOAP file link: You're linking to a GitHub webpage, not the file itself. Please use a direct link ("Raw" button), like this: https://raw.githubusercontent.com/vsevex/whixp/refs/heads/main/doap.xml
2. If a DOAP link is supplied, both `platforms` and `url` keys must be left empty. because those infos should be provided via DOAP file.

So in conclusion, your entry should look like this:

```
    {
      "categories": ["library"],
      "doap": "https://raw.githubusercontent.com/vsevex/whixp/refs/heads/main/doap.xml",
      "name": "Whixp",
      "platforms": [],
      "url": null
    },
```

guusdk:
Eek. Apologies for merging, I didn't know that.

Can we adjust the validation of PRs accordingly?

cal0pteryx:
No worries :) That's a good idea, I'll look into it.

vsevex:
Okay, I am opening a new PR now.

PR #1441, opened 2024-10-04
Translation for September newsletter

Millesimus:
Hey folks, here is my proposal for a German translation.


cal0pteryx:
Thanks a lot!

PR #1449, opened 2024-11-06
German newsletter translation

Millesimus:
Hey folks, here is my German translation for the XSF newsletter of October!


cal0pteryx:
Thanks!

PR #1453, opened 2024-11-11
Update Overview

butyoutried:
- Update some links.
- Remove ancient and irrelevant names of Pidgin and Jitsi.

There is a lot of unmaintained, severely outdated software on this page: Pandion, Soashable, SparkWeb, etc. Would it be better to move such projects to the "Other Software" section?


guusdk:
Maybe it's better to not explicitly name implementations on this page at all, as it makes for a bit of a messy read. It could instead refer to the DOAP-based tooling under the 'software' setting. Something along the lines of

> To find out what XMPP software supports \<FEATURE>, see \<LINK-TO-TOOL-WITH-PREDEFINED-FILTER>

(or better wording).

cal0pteryx:
I agree, it would be better to not explicitly name implementations on the Technology Overview pages. Not having dead links is a good start though, since this page needs a rework anyways.

PR #1467, opened 2024-12-26
Update instant-messaging.md

Mailfence-team:
An update on the current number of users and the relevant URL.


cal0pteryx:
Thanks!

PR #1471, opened 2025-01-09
German translation 2024/12

Millesimus:
Hey folks,

here is my German translation for the latest newsletter.


cal0pteryx:
Thanks a lot!

PR #1472, opened 2025-01-16
Extensions: Show <supersededby> specs in extension table

cal0pteryx:
This adds a button with a popover for all XEP in https://xmpp.org/extensions which have a `<supersededby>` tag set. 

All `<spec>` elements within the `<supersededby>` are rendered inside the popover. 

If possible, specs are transformed to an HTML link (for XEPs and RFCs).

Depends on xsf/xeps#1418

Example:
image

@iNPUTmice

### DIFF
--- a/themes/xmpp.org/layouts/shortcodes/xeps-table.html
+++ b/themes/xmpp.org/layouts/shortcodes/xeps-table.html
@@ -106,6 +106,7 @@
             </thead>
             <tbody>
                 {{- if .Site.Data.xeplist -}}
+                    {{- $xeplist := .Site.Data.xeplist -}}
                     {{- range sort .Site.Data.xeplist ".number" -}}
                         {{- $number_str := printf "%04g" .number -}}
                         {{ $last_revision_date := .last_revision_date | time }}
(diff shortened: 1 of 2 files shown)


iNPUTmice:
Hi,

thank you for that. What I had envision - but not communicated - was to make it way more obvious than that:

Underneath Message Archiving it would say:

> *Superseded-by*: XEP-0313: Message Archive Management

The idea was that if someone is searching for a functionality rather than a specific number they would almost always click on the big, obvious superseded-by link instead of on the XEP itself.

However I’m obviously not the only person that gets to decide this. Maybe we can limit the aforementioned display method to super obvious, clear cut cases. When the XEP is deprecated or deferred and the superseded points to something we can resolve and keep the method from this PR as a fallback for the less obvious ones. (Meaning in your specific example it would probably still show what your example shows because 295 is 'Active'.

And/or we can start by merging this PR and see where this leads us.

cal0pteryx:
Thanks! Yes, that's actually a better idea than my first approach. I think it should be obvious now if you look at this screenshot:
image

I added the title as well (at least for XEPs).

PR #1479, opened 2025-02-03
text-nowrap on subscribe/archives links for mailinglists

### LINKED ISSUE #1478: Mailinglist page has weird line-breaks
The web page that shows the mailing lists has some weird line-breaks in them. Perhaps using a `nowrap` directive for the sections that point to mailinglist archives would improve things.

Image

guusdk:
Introduces a shortcode for the block that contains the `(subscribe | archives)` links for each mailing list. This intends to prevent odd line-breakage within those blocks.

fixes #1478

Changes render as shown below:

image

### DIFF
--- a/content/community/mailing-lists.md
+++ b/content/community/mailing-lists.md
@@ -9,19 +9,19 @@ There are many active discussion venues for Jabber/XMPP developers, system admin
 
 ### For End Users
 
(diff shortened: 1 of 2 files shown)


cal0pteryx:
Thanks!

PR #1488, opened 2025-02-17
newsletter| TRANSLATE 2025-01 to German.

Millesimus:
Hey folks, here is my translation for the current newsletter.

I'm not quite sure whether I grasped the meaning of what you meant with ralphm's "swag" and Edwin's "dinner form", so it would be nice if you could double-check. :)


cal0pteryx:
Thanks a lot!

PR #1494, opened 2025-02-26
Update people

ralphm:


guusdk:
Have you considered if where you replaced "psa, the ED" with "psa" it would be more appropriate to replace "psa, the ED" with "mattj, the ED" ?

ralphm:
I did, but probably Peter is still the more knowledgeable person about both the founding documents and the trademark policy, so putting MattJ in the middle seems superfluous.

cal0pteryx (approved):
I tested it yesterday, lgtm!

PR #1496, opened 2025-03-06
Rename mierenhoop/xmpp DOAP to picomemo

mierenhoop:
Hi, the project now only provides an OMEMO implementation and has re-branded accordingly.


guusdk:
Thanks! Can you re-arrange the entries please? Computer-says-no when entries are not ordered alphabetically by case-folded name.

PR #1506, opened 2025-04-05
Correct FOSSY CfP deadline in March newsletter

badrihippo:
The FOSSY website states it as April 28th, not 5th May: https://2025.fossy.us/call-for-proposals/


cal0pteryx:
Thanks!

PR #1509, opened 2025-04-09
Fix links in hosted DOAPs

butyoutried:


guusdk (approved):
Thanks
- all new URLs work
- The change for libstrophe results in a DOAP similar to that of strophe.js, which seems to be desirable to me.
- the old URLs for isode.com work, but redirect to the new URLs. Using the new URLs directly seems fine to me.

cal0pteryx:
Thanks

PR #1510, opened 2025-04-09
Overview: an incomplete cleanup

butyoutried:
Somebody should rework this page anyway. Let's start with some sort of cleanup.

- Unmaintained (for a long time) projects are moved to the Other Software.
- Unmaintained projects that are already included in the Other Software are deleted from the Overview.
- The links are fixed.

### DIFF
--- a/data/software.json
+++ b/data/software.json
@@ -1042,17 +1042,6 @@
             "client"
         ],
         "doap": null,
-        "name": "JWChat",
-        "platforms": [
-            "Browser"
-        ],
-        "url": "https://github.com/sstrigler/jwchat"
-    },
-    {
-        "categories": [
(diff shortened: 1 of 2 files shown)


cal0pteryx:
Thanks for your efforts!

Unmaintained software should generally move to `software_archive.json`, so it does not clutter the other list, since it's already pretty large. Could you move the unmaintained ones there?

cal0pteryx:
Thanks!

> Done. I will move other discontinued or completely abandoned (with 10-15+ years of inactivity or so) projects from the software.json to the software_archive.json then. Is that OK?

Yes, that would be great! I merged this PR, so these changes can be done in a separate PR.

PR #1513, opened 2025-04-13
German translation for the open letter to Meta

Millesimus:
Hey everyone,

here is my translation of the open letter to Meta. Please double-check, it was quite the piece and I translated same passages more freely than I usually do with the newsletters.

### DIFF
--- a/content/announcements/open-letter-meta-dma/technical-briefing.de.md
+++ b/content/announcements/open-letter-meta-dma/technical-briefing.de.md
@@ -1,0 +1,284 @@
+---
+title: "Technische Zusammenfassung: Argumente für XMPP - Warum Meta sich echte Messenger-Interoperabilität bemühen muss"
+layout: single_open_letter_meta_dma
+date: 2025-03-27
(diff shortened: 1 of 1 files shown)


dwd:
Thanks! Can you add a disclaimer to the effect of "this is a community-provided translation of the open letter sent to Meta for the convenience of German speakers, and is not itself an official communication of the XSF" or similar? I think it's great you've translated it to maintain its sense rather than strictly keeping to a word-for-word translation, but I also wouldn't want the German press, for instance, to assume every phrase is what we sent Meta and the EU.

Echolon:
All translations show an automatic notification on the top, see this: https://xmpp.org/de/2025/04/der-xmpp-newsletter-f%C3%BCr-m%C3%A4rz-2025/

14 abr 2025 9:46:52 Dave Cridland ***@***.***>:

dwd:
Ah-ha! I'm essentially monoglot-English, so I'd never noticed before.

Echolon (approved):
Many thanks!

PR #1518, opened 2025-04-27
WIP: Xsf announce

bkmgit:


Echolon:
@bkmgit many many thanks! I'm very happy to see what you do!

Some points:

* I think you are pushing content from your first PR (like I see the banners again in the commit). Is that intended?
* Was there are reason that you made the old announcement as a logo? In general, I would just like to use your new design
* Though, I mis-communicated. I only need the banners, no logos, but I'm fine to keep what you made.
* For XSF I think the font is not the best choice. You don't need to mimic it, but the one you chose is a bit odd.
* So, maybe the XMPP Market I was also looking for a banner. It can be also in the design you mimic-ed so far (like the newsletter. Some design suggestions below
** Screenshot_20250430_225649
** Move the XMPP Logo into the stall header / top
** Have the icons below and the person alone
** Have less icons, maybe join the two left.

Awesome work again! <3

bkmgit:
@Echolon XMPP Market banner is similar to the other banners now.  Further suggestions welcome.

@Echolon  @badrihippo @cal0pteryx  XSF logo probably needs some discussion. For example the font choice for XSF. Do we also want to use a single icon for Announcements?

cal0pteryx:
Yes, announcements should use the same icons imho. Best would be to unify it with the icon used in xmpp_announcement_sticker.svg

bkmgit:
> > @Echolon There was not a clear decision on the font.
> > ```
> > * Do we want to remove the XSF Announcment?
> > 
> > * Do we want to keep XSF Announcment and use [Librestile](https://github.com/ocelothe/Librestile)?
> > 
> > * Do we want to keep XSF Announcment and use the font used in the text XMPP which is [Eurostile Extended Bold](https://www.myfonts.com/products/bold-extended-2-63893-eurostile-lt-356389/licenses) and requires someone with a desktop license to provide the text if we also want to use it for the text `XSF`.
> > ```
> 
> Hey,
> 
> so I think all is clarified:
> 
>     * We dont need to remove the font. We can use it again.
> 
>     * So, no need for another style, right?

We do not have XSF using [Eurostile Extended Bold](https://www.myfonts.com/products/bold-extended-2-63893-eurostile-lt-356389/licenses)  the current text in XSF uses [Librestile](https://github.com/ocelothe/Librestile)

The main use is in https://github.com/xsf/xmpp.org/pull/1518/files#diff-a7b5f1f88a231848548d3b4af01660d34a6545ef7d3cb2edf35196bd87e3d985

> 
>     * No license required. Just use what's in the XSF Logo

Suggested:

https://github.com/xsf/xmpp.org/pull/1518/files#diff-c1877a622f14a0370593a094d80606c92b683a48c4bde44509a4d51dbb46554e

though can leave out for now as it seems it will not have much use elsewhere, and
someone may come up with something better or at least go through a community
vetting process.

Echolon:
> > @bkmgit many thanks! But with this the concerns are gone now?
> 
> Text style does not quite match. It would probably be best for XSF as a legal entity to acquire a desktop license for the font [Eurostile Extended Bold](https://www.myfonts.com/products/bold-extended-2-63893-eurostile-lt-356389/licenses) - but will leave it to someone else to pursue this as text mismatch is less of a concern than use of a font without a license or a modified font with unclear origin/unattributed origin.

Ok thanks. My question is if you could just only use the Librestile font for all images?

PR #1521, opened 2025-05-06
Fix Movim mobile screenshot in April 2025 newsletter

badrihippo:
The actual image is uploaded as `.jpg`, not `.png`

Also, I'm assuming "movile" is a typo and not some fancy way of saying "Movim mobile device"...although it could be that too ;-)


cal0pteryx:
Thanks!

PR #1522, opened 2025-05-10
German translation for the April newsletter

Millesimus:
Hey everyone,

here is my translation for the April newsletter. :)


cal0pteryx:
Thanks!

PR #1542, opened 2025-08-06
software.json: remove Psi+ as it's just a beta build of Psi

stokito:
The Psi+ long time ago was a fork of Psi but now this project merged and the Psi+ is called by devs as "nightly build". The Psi installable from all distros is built from the Psi+

### DIFF
--- a/data/software.json
+++ b/data/software.json
@@ -1265,15 +1265,6 @@
     },
     {
         "categories": [
-            "client"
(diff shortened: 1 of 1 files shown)


cal0pteryx:
@Neustradamus since this has been rejected already in the ticket you've linked, I urge you to stop.

PR #1546, opened 2025-08-11
Italian translations of the June and July 2025 Newsletters

resoli:
As said in https://github.com/xsf/xmpp.org/pull/1545#issue-3309958871 the two files mention links to xmpp-it and nicola's  collection of newsletters.


cal0pteryx:
Thanks!

PR #1547, opened 2025-08-15
Fix movim account link

nemmig:
Now all previous links to the Movim account, that I could find, point to the right direction: https://mov.im/community/news.xmpp.org/Newsletter

### DIFF
--- a/content/blog/newsletter/2025-06-10-newsletter.de.md
+++ b/content/blog/newsletter/2025-06-10-newsletter.de.md
@@ -106,7 +106,7 @@ Letzte Aufrufe (_Last Calls_) erfolgen sobald sich der Status eines XEPs konsoli
 Teile diese Neuigkeiten auch in anderen Netzwerken:
 
 - [Mastodon](https://fosstodon.org/@xmpp/)
-- [Movim](https://mov.im/blog/xmpp-official)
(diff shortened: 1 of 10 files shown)


Echolon:
For the moment I prefer to just have one feed for everything. We don't have the capacity to add more feeds we need to maintain.

edhelas:
@nemmig @Echolon Feel free to name it wherever you want, you should know that the name of the node is CaseSentitive.

And yes I'd suggest to put everything on the Community and leave the commteam@xmpp.org posts empty, you can also directly publish on the Community using your XMPP account if you put it as a Redactor :) This is how I do it for the Movim Blog: https://mov.im/community/pubsub.movim.eu/Movim

Echolon:
I see I need to get used to it.

Tough, why do one not see the post I made here?  https://mov.im/community/news.xmpp.org

nemmig:
@Echolon Just for the record: I can s/Newsletter/Blog in all those files no problem if you want me to.
Just ask me and I'll do it in a couple of minutes.
Will also fix the wiki entries.

cal0pteryx:
What's the status here?

nemmig:
D'oh! I only wanted to remove the diff from my PR, not the whole darn file!

PR #1560, opened 2025-09-16
German newsletter translation July 2025

Millesimus:
Hey folks,

here is my translation for the July newsletter. Sorry that it took me so long, I try to be a little faster with the August edition.

Cheers,

Milles


cal0pteryx:
Thanks!

PR #1591, opened 2025-11-04
Activate event: XMPP Summit 28

Echolon:


data/announcement.json
```diff
@@ -1,5 +1,5 @@
 {
-    "enabled": false,
-    "text": "Join the XMPP Sprint Berlin, 23-25 May",
-    "url": "https://wiki.xmpp.org/web/Sprints/2025-05_Berlin"
+    "enabled": True,
```
cal0pteryx: There is a comma missing

cal0pteryx: The diff gave a wrong impression. It was the boolean which needs to be lowercase.

Echolon:
Thanks!

PR #1593, opened 2025-11-06
Optimize date readability in announcement.json

Echolon:


data/announcement.json
```diff
@@ -1,5 +1,5 @@
 {
     "enabled": true,
-    "text": "XMPP Summit 28 - 29th & 30th Jan 2026",
+    "text": "XMPP Summit 28  ||  29th - 30th Jan 2026",
```
cal0pteryx: Double-barrel seems arbitrary for separating title and date. How about this:

```suggestion
    "text": "XMPP Summit 28 - Join us in Brussels, 29th - 30th Jan 2026",
```

PR #1672, opened 2026-05-02
Use shared data source for sponsors & apply sponsor status changes

guusdk:
This switches to using a common data source, rather than hard-coding data for sponsors on various pages. From this data, the footer and page that lists all sponsors is populated.

The page that lists all sponsors is modified to differentiate between the various levels of sponsors.

Only the longer descriptive texts for gold sponsors are not included in the re-usable data.

### DIFF
--- a/content/community/sponsors/_index.md
+++ b/content/community/sponsors/_index.md
@@ -6,38 +6,8 @@ aliases:
   - "/community/sponsors.html"
 ---
 
-The XSF is generously supported by the sponsors below, please visit their xmpp.org pages (linked below) and sites to say thank you.
+The XSF is generously supported by the sponsors below. Please visit their pages and websites to say thank you.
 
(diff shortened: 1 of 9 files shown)


guusdk:
I've now included changes to Extensible Messaging Ltd, Tigase and Tygrys sponsorship statuses to this PR.

The resulting sponsor list now looks like this:

image

cal0pteryx:
Hi @guusdk and thank you for this work! I've restructured and generalized the shortcode a bit in https://github.com/xsf/xmpp.org/commit/4cba89977e038c39c20c9d108adafd5c390642ea

On top of that, I added a "website" key to the data, which I'm currently using for the silver sponsors listing (on logos).

guusdk:
Those are solid improvements, thanks!

PR #1673, opened 2026-05-03
Software: Update jabber.el DOAP and homepage URLs to Codeberg

thanosapollo:


nemmig:
Updated jabber.el DOAP URL to Codeberg
Thank you for pointing it out!

PR #1680, opened 2026-05-27
Add Teletyptel 2.0 to software directory

edwtie:
Adds Teletyptel 2.0 as a web-first XMPP client entry.

Teletyptel 2.0 is an accessible realtime communication client focused on standard chat and XEP-0301 real-time text, with Android and iOS packaging planned from the same web client direction.

Validation:

- `python ./src/lint_software_list.py ./data/software.json`

### DIFF
--- a/data/software.json
+++ b/data/software.json
@@ -1621,6 +1621,19 @@
     },
     {
         "categories": [
+            "client"
+        ],
+        "doap": null,
+        "name": "Teletyptel 2.0",
+        "platforms": [
+            "Android",
+            "iOS",
+            "Web"
+        ],
+        "url": "https://github.com/edwtie/Tiedragon.XmppMessenger"
+    },
+    {
+        "categories": [
             "server"
         ],
         "doap": "https://raw.githubusercontent.com/tigase/tigase-server/master/tigase-server.doap",


guusdk:
Thanks for the submission.

After reviewing the repository, this project appears to still be at a concept/prototype stage rather than a usable client release. There are currently no releases, no public instance or demo, very limited development history, and no website or user-facing documentation. The PR text also describes key platform support as planned work.

Since xmpp.org is used by visitors to discover software they can realistically evaluate and use today, I do not think the project is ready to be listed yet.

Please feel free to resubmit once there is a usable release and sufficient documentation for end users.

edwtie:
Thanks for the clear review.

I agree with the concerns. The first submission was too early and the project looked more like a concept than something people could evaluate.

I have now addressed the main points:

- Added a first public prerelease: v0.1.0-alpha1
  https://github.com/edwtie/Tiedragon.XmppMessenger/releases/tag/v0.1.0-alpha1

- Added a downloadable Alpha 1 web demo package:
  teletyptel-0.1.0-alpha1-web-demo.zip

- Reworked the README so it describes what can be evaluated today instead of only planned work.

- Added end-user and reviewer documentation:
  - Getting Started
  - User Guide
  - Alpha 1 Release Notes
  - Project Page notes
  - XSF readiness checklist

- Added a public license and changelog.

This is still clearly marked as an alpha/pre-release. It is not presented as a finished production messenger, but it now has a repeatable local demo path and release package that reviewers can run.

I understand that a public hosted demo instance may still be expected before listing. If that is required, I will wait with resubmission until that is online too.

Thanks again for the feedback; it helped make the repository much clearer.

PR #1675, opened 2026-05-15
Add Ethora to software list

phwizard:
Adds [Ethora](https://ethora.com) to `data/software.json`.

Ethora is an open-source XMPP-based chat & messaging platform with SDKs for React, native iOS/Android, and WordPress, plus an AI agent / chatbot framework. The server side is built on ejabberd.

- Categories: `integrated-platform`, `library`
- Platforms: JavaScript, Android, iOS
- URL: https://ethora.com
- Source: https://github.com/dappros/ethora (Apache-2.0)
- DOAP: not yet published — happy to follow up with one if helpful.

Inserted alphabetically between `Escalus` and the next entry.

PR #1681, opened 2026-05-27
Ethora: link to upstream DOAP file

phwizard:
Follow-up to #1675 (Add Ethora to software list).

The Ethora repo now ships a DOAP file at https://raw.githubusercontent.com/dappros/ethora/main/doap.xml, so this fills in the previously-`null` `doap` field on the Ethora entry in `data/software.json`.

Per @guusdk's suggestion in #1675, this should improve how Ethora is rendered on xmpp.org.

cc @cal0pteryx @guusdk

### DIFF
--- a/data/software.json
+++ b/data/software.json
@@ -488,7 +488,7 @@
             "integrated-platform",
             "library"
         ],
-        "doap": null,
+        "doap": "https://raw.githubusercontent.com/dappros/ethora/main/doap.xml",
         "name": "Ethora",
         "platforms": [
             "Android",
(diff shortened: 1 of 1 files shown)


guusdk:
Hi, thanks for that!

I strongly suggest you add a logo (`<schema:logo rdf:resource="https://example.org/mylogo.svg"/>`) as that will immediately help stand out your project!

As the build automation says, you'll need to remove some information from the metadata when you provide a DOAP. That's to prevent having to (potentially conflicting) sources of truth.

By the way, there no need to @-mention us, we get notified anyways.

guusdk (approved):

phwizard:
thank you Guus!

PR #1685, opened 2026-07-03
Jabber: Update Trademark Enforcement Agreement document

cal0pteryx:
@stpeter all three documents regarding the Jabber trademark are already available at https://xmpp.org/about/xsf/jabber-trademark/background/#4-governing-agreements

Since your version of the Trademark Enforcement Agreement document seems to have some more pages, I've replaced the one published on xmpp.org.with the updated version.


guusdk:
Merging. For the record, on what this changes: the previously published Trademark Enforcement Agreement was a two-page scan containing the agreement body and signature page only. The agreement's recitals reference "a Jabber Trademark Position Statement ('JTPS'), a copy of which is attached hereto as Exhibit A" but that exhibit was not part of the published scan.

This updated scan is the complete document: the same two pages (identical text, same signature dates), plus Exhibit A: the Jabber Trademark Position Statement itself, which sets out the trademark usage guidelines the agreement refers to.

Thanks @stpeter for digging these out.

ge0rg:
Exhibit A is attached to the older scan in `static/docs/Trademark_Letter_Of_Intent.pdf`. In fact, the two older scans are using monochrome with a lower resolution, so I would propose replacing them as well. Thoughts?

PR #1687, opened 2026-07-19
Add Livewire (Ubuntu Touch Client)

ProjectMoon:
I've been working on an XMPP client for Ubuntu Touch, and I now think it's ready enough to be on the list of clients.

### DIFF
--- a/data/software.json
+++ b/data/software.json
@@ -978,6 +978,15 @@
     },
     {
         "categories": [
+            "client"
+        ],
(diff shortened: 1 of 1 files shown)


cal0pteryx:
Looking at the CI pipeline (the "Prepare Software List" step), it looks like your project's DOAP file is not accessible to the CI. This may be due to bot-prevention mechanisms present at the server where you host the file.

ProjectMoon:
I updated the Anubis configuration. Should be accessible via HTML and raw XML now.

PR #1699, opened 2026-08-15
Events: Move to past events after end_date not start_date

PolarianDev:
Currently events are moved to `past_events` during the conference as the past events are checked against the `start_date` not `end_date`.

This patch fixes this by checking against the `end_date` instead

### DIFF
--- a/data/events.json
+++ b/data/events.json
@@ -148,7 +148,7 @@
         "title": null,
         "location": "Portland (US)",
         "date_start": "2023-05-30",
-        "date_end": null,
+        "date_end": "2023-05-30",
         "url": "https://2023.fossy.us/",
         "desc": "around Fossy"
     },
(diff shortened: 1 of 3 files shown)


PolarianDev:
This is currently not ready to merge, there is a edge case where if its the final day (aka if FrOSCon ended on the 15th) it will match against that.

I assume its because when comparing the time, it compares against the time at midnight on the day, and because its hours into the day, its now less than, and thus moves along.

I haven't a clue how to turn the "now" into a "today" but 23:59 and then the lt the same, and that would fix it, but again no clue :p

cal0pteryx:
I've pushed a fix :) We now take the date of `now` for comparison, which should make it more robust overall.

cal0pteryx:
Thanks a lot, @PolarianDev !

PR #1701, opened 2026-09-02
Add XMPP Leaflet (a5) to images/promo

guusdk:
This is the EPS file for the A5 flyer (white and blue) that describes XMPP.


cal0pteryx:
Thanks!

PR #1702, opened 2026-09-03
Update software.json to add MirrorFly to xmpp

alexsam9866:


alexsam9866 (commented):
Request to add mirrorfly to database

guusdk:
Hi Alex! Thanks for this! A quick check on eligibility: Is XMPP a substantive part of MirrorFly, rather than merely something used in the by-lines?

Also make sure that the file that you edit keeps being conformant to the validation rules that exist for this file. You can read about those here: https://github.com/xsf/xmpp.org/blob/master/src/README.md

alexsam9866 (commented):
Have updated the order issue. And MirrorFly's chat channels are built as an XMPP protocol.

cal0pteryx:
From a quick glance at the website I think your product uses XMPP for its chat functions internally? This would make it a better fit in https://xmpp.org/uses/instant-messaging/ I think.

PR #1709, opened 2026-09-16
Update dark theme for compatibility with non-JS users

entibo:
**Problem**: dark mode styles are controlled by `[data-bs-theme=dark]` for the Bootstrap theme, and a media query for `styles.css`. With **Javascript disabled**, only the latter are applied which results in a mix of light and dark styles (**unreadable in places**).

**Fix**:  Apply dark mode styles using the data attribute exclusively. No-JS users get light mode only as this is how Bootstrap is setup in this project. It changes nothing for everyone else.


cal0pteryx:
Thanks!

